PR #53, opened 2019-05-08
[Registration] Add multistep registration form

paiva:
This PR:
* Modifies the current registration form to be a multi-step process (3 steps)
* The JS is still breaking, needs some work


shots47s:
Has this pull request been replaced by @DESm1th's work or do we still need to look at finishing this pull request?

PR #29, opened 2019-04-15
[Pipelines] Add pipeline react element

paiva:
This PR:
* [ ] Adds the PipelineElement react component to visualize pipelines on `/pipelines`

Still a work a progress

PR #55, opened 2019-05-28
Integrate boutiques pipeline tool

joeyzhou98:
In tools and pipelines tab, the backend needs a module/package to retrieve and search for pipelines. Boutiques is a good choice for this task.


app/routes.py
```diff
@@ -390,3 +394,21 @@ def forums():
...
+                os.makedirs(".cache/boutiques")
+            thr = UpdatePipelineData()
+            thr.start()
+            thr.join()
+        elif time.time() - os.path.getmtime(".cache/boutiques/pipelineData.json") > 300:
+            UpdatePipelineData().start()
+
+        #search cache and return result
+
+        return json.dumps("{}")
```
glatard: You could dump the whole content of `pipelineData.json` in case no search query is provided, as is the case in `bosh search`.

app/threads.py
```diff
@@ -0,0 +1,11 @@
...
+from boutiques import search
+import json
+
+
+class UpdatePipelineData(Thread):
+
+    def run(self):
+        with open(".cache/boutiques/pipelineData.json", "w") as f:
+            data = search("-m 1000", "-nt")
+            json.dump(data, f, indent=4)
```
glatard: You should also pull all the descriptors and use the info inside to provide a more detailed view if the user clicks on one.

joeyzhou98:
Hi @glatard, here is a fully functional and commented integration of boutiques in the CONP-Portal!
There are a few points I want to bring up:

- The code checks whenever receiving a request if the cache data is older than 5 minutes (could make it an environment variable or change hard-coded value) and if it is the case, will pull all the descriptors similar to `bosh search` and will perform a `bosh pull` on every single descriptor on the list and store them in the cache. The problem is that `bosh pull` will first check in the cache if the specific descriptor file is there first and if it is, it won't pull it:
 ```
 def pull(self):
    # return cached file if it exists
    if os.path.isfile(self.cached_fname):
        if(self.verbose):
            print_info("Found cached file at %s"
                       % self.cached_fname)
        return self.cached_fname
```
   This might not fully update those specific descriptors as we don't know if someone pushed to them 
   while those files are cached on the server so I am thinking maybe if the cache is older than 2 hours, 
   fully reset the cache with fresh data by deleting the .cache/boutiques directory and pulling again. Let 
   me know if this is a possible enhancement.


- @paiva the payload currently has the format of [payload.txt](https://github.com/CONP-PCNO/conp-portal/files/3235293/payload.txt), let me know if there is any change in the structure or anything that will make rendering it on the frontend easier.


- The current search mechanism only makes a
 `if search_query in descriptor["TITLE"] + descriptor["DESCRIPTION"]:`
  Let me know if you guys have a better mechanism for searching in mind

**Comment by glatard** [comment]
Hi @joeyzhou98, it looks great!
* Descriptors are immutable: pushing a new version of a descriptor results in a new Zenodo id. So I think that the current behaviour of bosh pull is correct, no need to flush the cache.
* The search could be more general: `if search_query in str(descriptor)`. That would include tags and any other field.
I will review the code later and provide feedback.

File: app/routes.py [hunk 3]
```diff
@@ -390,3 +394,58 @@ def forums():
...
+        with open(all_desc_path, "r") as f:
+            all_descriptors = json.load(f)
+
+        #fetch every single descriptor
+        descriptors = []
+        for descriptor in all_descriptors:
+            file = descriptor["ID"].replace(".", "-") + ".json"
+            file_path = os.path.join(cache_dir, file)
+            with open(file_path, "r") as f:
+                descriptors.append(json.load(f))
```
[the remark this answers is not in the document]
joeyzhou98: > Hi @joeyzhou98, this is a very great first PR! See my comments in the code. I also installed your branch and tried it; search actually didn't show anything:
> Screenshot from 2019-05-31 14-00-35

I have made the requested data available to the frontend but more work is needed to display them. I think @paiva is handling the frontend. We can talk more on Monday.

joeyzhou98:
@glatard let me know if there is anything else to change

app/routes.py
```diff
@@ -390,3 +399,47 @@ def forums():
...
+        cache_dir = os.path.join(os.path.expanduser('~'), ".cache", "boutiques")
+        all_desc_path = os.path.join(cache_dir, "all_descriptors.json")
+        all_detailed_desc_path = os.path.join(cache_dir, "detailed_all_descriptors.json")
+
+        #fetch data from cache
+        with open(all_desc_path, "r") as f:
+            all_descriptors = json.load(f)
+
+        with open(all_detailed_desc_path, "r") as f:
+            detailed_all_descriptors = json.load(f)
```
glatard: Isn't there a way to keep that in memory? Now you're going to reload all these files with every query.

app/routes.py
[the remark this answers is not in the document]
glatard: Having a basic client implementation would be useful, do you think you could write one starting from @paiva's code at https://github.com/CONP-PCNO/conp-portal/pull/29? You could pull that branch here. We can also discuss that today with @paiva.

app/routes.py
joeyzhou98: From my knowledge, if you want to use threads to update data in the background in memory, then there is no other choice than to declare those data variables global which is bad practice because I ran a few tests and I couldn't pass the data variables to the threads to be updated as the threads create local copies of those variables and update those instead and there is no way to return those updated data variables as we are using threads except via caching in files to be read.

I think we have few options to go with:
- Either keep the code like this, meaning we will have to read from the cache everytime there is a query, but at least we can use threads to update data in the background and there is no global variable
- We can stop using threads meaning that the query that triggers the data update will take a little bit longer to be responded to since it will have to wait for the updateData() method until the local data is fully updated.
- Setup some kind of database we can query from and that is self updated/updated by threads, but this options is pretty similar from the first option as they both read/query from a third party unless I am wrong.

Let me know if there is any better options

glatard: Option 1 seems the best for now. Option 3 is definitely the way to go when we want to improve performance. Thanks for looking into this!

glatard:
Hi @joeyzhou98, let me know when this is ready for another round of testing!

joeyzhou98:
Hello @glatard, the data should now show very minimally when you load the page. More work is needed to beautify the page, but I really don't think I am the one to do it as it is out of my area of expertise.

And I flattened the "short_descriptor" and "long_descriptor" pairs in each elements so the payload should look like [this](https://github.com/CONP-PCNO/conp-portal/files/3283639/payload.txt) now, @paiva.

Let me know if there are any questions!

glatard:
Hi @joeyzhou98, I'm excited to test this, but when I run it, I get the following error:
```
127.0.0.1 - - [12/Jun/2019 22:27:51] "GET /static/css/font-awesome.css HTTP/1.1" 304 -
127.0.0.1 - - [12/Jun/2019 22:27:51] "GET /static/css/font-awesome.min.css HTTP/1.1" 304 -
127.0.0.1 - - [12/Jun/2019 22:27:52] "GET /static/img/neurolibre-icon.png HTTP/1.1" 304 -
127.0.0.1 - - [12/Jun/2019 22:27:52] "GET /pipeline-search?cursor=0&limit=10&search=&sortComparitor=asc&sortKey=title HTTP/1.1" 500 -
Traceback (most recent call last):
  File "/home/glatard/venvs/conp-portal/lib/python3.7/site-packages/flask/app.py", line 2309, in __call__
    return self.wsgi_app(environ, start_response)
  File "/home/glatard/venvs/conp-portal/lib/python3.7/site-packages/flask/app.py", line 2295, in wsgi_app
    response = self.handle_exception(e)
  File "/home/glatard/venvs/conp-portal/lib/python3.7/site-packages/flask/app.py", line 1741, in handle_exception
    reraise(exc_type, exc_value, tb)
  File "/home/glatard/venvs/conp-portal/lib/python3.7/site-packages/flask/_compat.py", line 35, in reraise
    raise value
  File "/home/glatard/venvs/conp-portal/lib/python3.7/site-packages/flask/app.py", line 2292, in wsgi_app
    response = self.full_dispatch_request()
  File "/home/glatard/venvs/conp-portal/lib/python3.7/site-packages/flask/app.py", line 1815, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/home/glatard/venvs/conp-portal/lib/python3.7/site-packages/flask/app.py", line 1718, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/home/glatard/venvs/conp-portal/lib/python3.7/site-packages/flask/_compat.py", line 35, in reraise
    raise value
  File "/home/glatard/venvs/conp-portal/lib/python3.7/site-packages/flask/app.py", line 1813, in full_dispatch_request
    rv = self.dispatch_request()
  File "/home/glatard/venvs/conp-portal/lib/python3.7/site-packages/flask/app.py", line 1799, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/home/glatard/code/conp-portal/app/routes.py", line 420, in pipeline_search
    with open(all_detailed_desc_path, "r") as f:
FileNotFoundError: [Errno 2] No such file or directory: '/home/glatard/.cache/boutiques/detailed_all_descriptors.json'
```

paiva:
@joeyzhou98 Is it still giving errors? Let me know when I can test it on my side

joeyzhou98:
> @joeyzhou98 Is it still giving errors? Let me know when I can test it on my side

@paiva I need to make enhancements to the frontend display of pipeline elements. I will let you know when I am done, but you can still download my branch and test it, it will however be a very minimalistic frontend display of the data

joeyzhou98:
Hello, @glatard, the frontend should be a little better now! With all the requested links attached to title, id and logos.

@paiva feel free to test it now!

glatard:
Hi @joeyzhou98, great work! Minor comments:
* Search should be case-insensitive: currently, searching for `bids` doesn't return anything while `BIDS` does.
* Sorting by downloads should be in descending order: most downloaded tools first. 
* In fact I don't think that "Sort by" makes much sense: you could always sort by downloads and remove that box.
* We should add a paragraph to explain how to run pipelines locally, between the page title and the search box, for instance: "CONP pipelines can be easily installed and run on any computer with a container engine. For more information see our tutorial [here](https://nbviewer.jupyter.org/github/boutiques/tutorial/blob/master/notebooks/boutiques-tutorial.ipynb#reusing_tools)."

joeyzhou98:
> Hi @joeyzhou98, great work! Minor comments:
> 
> * [ ]  Search should be case-insensitive: currently, searching for `bids` doesn't return anything while `BIDS` does.
> * [ ]  Search should operate on tags too: for instance `bosh search brainlife` returns a tool not returned by the portal.
> * [ ]  Search should operate on tool name too: for instance `bosh search example` doesn't return `zenodo.2644621` while `bosh search` does.
> * [ ]  Tools that have non-empty `online-platform-urls`, for instance `zenodo.2644621` should have a link to their online platform.
> * [ ]  Sorting by downloads should be in descending order: most downloaded tools first.
> * [ ]  In fact I don't think that "Sort by" makes much sense: you could always sort by downloads and remove that box.
> * [ ]  We should add a paragraph to explain how to run pipelines locally, between the page title and the search box, for instance: "CONP pipelines can be easily installed and run on any computer with a container engine. For more information see our tutorial [here](https://nbviewer.jupyter.org/github/boutiques/tutorial/blob/master/notebooks/boutiques-tutorial.ipynb#reusing_tools)."

1) Done
2) Done
3) Searching `example` didn't return `zenodo.2644621` was caused by the tool name having `Example` with capital "E". Fixed with 1.
4) It is already linked via the globe icon on the right. Let me know if you want the link to be place elsewhere
5) Done
6) Unfortunately, the "Sort By" box/feature is in a React component that I am reusing from the Search portal page which means if you want it to be removed, then I will need to duplicate all the components involved just for the Pipeline page. I have changed the "Sort By" options to "Downloads: High to Low" and "Downloads: Low to High". Let me know if you still want to have the "Sort By" feature removed
7) Done

glatard:
Hi @joeyzhou98, awesome work! I made a few minor fixes in a PR to your branch. Once you merge this, I think this is ready for @paiva's review!

glatard:
@paiva, would you have a few minutes to look at this?

paiva:
I know. I will get to it

paiva:
@joeyzhou98 Amazing work. Seriously.

glatard:
Thanks for reviewing this @paiva.

PR #65, opened 2019-07-01
Update temporary path for zenodo pipelines

paiva:
@glatard @joeyzhou98 I need to change the path of these zenodo json files, momentarily. When I deploy the app on Heroku, it cannot find this `.cache` directory. So, I have commented out the `cache_dir` and updated it with a path inside `/static/pipelines/`. This is a hack to get the demo ready for Tuesday, this will no longer be necessary when we finish migrating to a production ubuntu server (it is currently in progress), where a `.cache` directory will be available.

### DIFF
--- a/app/routes.py
+++ b/app/routes.py
@@ -409,7 +409,10 @@ def pipeline_search():
         # initialize variables
         search_query = request.args.get("search").lower()
         sort_key = request.args.get("sortKey") or "downloads-desc"
-        cache_dir = os.path.join(os.path.expanduser('~'), ".cache", "boutiques")
+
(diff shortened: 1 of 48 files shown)


paiva:
Yes, because of the way Heroku works. It can only read files that are
static on the app. For some reason, I am having trouble getting those files
when the app runs for the first time (this didn't happen on my Mac). I need
to investigate why. This is mainly to meet the demo requirements

glatard:
Then maybe you could leave them in the app dir but not commit them?

paiva:
I can't deploy what is not committed. Do these files change regularly? I
will think of another way

Issue #73, opened 2019-08-20
Create a distinct Development and Production configuration environments


shots47s:
## Purpose
In order to facilitate development and deployment, create the capability to distinctly set up a development vs. production environment.  

## Context
This way developers can set up a local env without the need to have use a dedicated VM.  Should ensure that the dev environment is deployable quickly on a laptop.

## Possible Implementation (optional)
using a well defined configuration definition for flask app, potentially multiple ones for each environment, and allowing a .flaskenv file to define local variables specific to the machine with flask-dotenv.

## Related issues (optional)
#71

PR #82, opened 2019-09-05
Loc #73

liamocn:
---
name: Create a distinct Development and Production configuration environments
about: In order to facilitate development and deployment, create the capability to distinctly set up a development vs. production environment.
title: Feature implementation #73
labels: enhancement
assignees: @liamocn 

---

## Checklist

- [ ] **DO** Unit tests pass.

no unit tests to run

## Purpose

In order to facilitate ease of deployment in different environments, a .flaskenv file is used in the project root. This file holds the environment variables for the project . Using production credentials vs development creds or setting the debug mode, etc is simply a case of using a production .flaskenv in place of the development .flaskenv 

## Current behaviour

Currently environment variables are configured using a mixed approach, some being hardcoded and other relying on global environment variables set within the developers own workspace

## New behaviour

The new behaviour is that the environment is configured from the one place, a .flaskenv file, any number of which can be used to service different environments, etc.
 
- [ ] Yes
- [X ] No

## Implementation Detail

This work uses python-dotenv to set environment variables from a single point, and anything that was hardcoded within the project has been moved to .flaskenv


config.py
```diff
@@ -5,6 +5,10 @@
 """
 import os
 
+from dotenv import load_dotenv
+
+basedir = os.path.abspath(os.path.dirname(__file__))
+load_dotenv(os.path.join(basedir, '.flaskenv'))
```
shots47s: I think this is actually automatic, dotenv looks for this file in the base directory.  But that being said, never hurts to be specific.

PR #84, opened 2019-09-10
Initial implementation of PyTest Framework

shots47s:
---
name: Added Unit Testing through the PyTest Framework
about: added tests directory
title: 'Feature implementation: Pytesting'
labels: enhancement
assignees: ''

---

## Related issues
None

## Checklist

- [x] **DO** Unit tests pass.
- [X] **DO** If new feature, created unit test.

## Purpose
Added a pytesting framework with some example Unit, Database and blueprint tests.

## Current behaviour
No testing

## New behaviour
pytest now runs a series of tests, and more can be created.
 
- [ X] Yes
- [ ] No

## Implementation Detail
There is a new directory, /tests.  Check out conftest.py for the fixtures that create and teardown the test build, it is pretty cool.  Pytest rules!


.travis.yml
```diff
@@ -10,5 +10,5 @@ install:
 - pip install -r requirements.txt
 
 script:
-- pycodestyle --max-line-length=99 app/*.py app/*/*.py
+- pycodestyle --max-line-length=99 -r *.py app/*.py
```
glatard: Shouldn't Travis be running `pytest` here?

glatard: You should also configure coveralls, to measure test coverage, and prevent merging PRs that lead to decrease in coverage. See example here: https://github.com/boutiques/boutiques/blob/master/.travis.yml#L53-57

shots47s: Good suggestion, doing so now.

tests/conftest.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+@pytest.fixture(scope='session')
+def app(request):
+    """
+    This is creates the mock app for testing, it uses the
+    TestingConfig in Config.py
+    """
+    app = create_app(config_settings=TestingConfig)
+    ctx = app.app_context()
```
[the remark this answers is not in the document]
shots47s: hmm, ok, I will give it a try, not sure why that was happening. It was passing for me, did you update your pip installs?  There are some new modules that need to be added, I should have said that.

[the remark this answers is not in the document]
liamocn: yes i updated dependencies. Strange it was passing for you but the travis builds were definitely failing too for the same reason I had.

.travis.yml
shots47s: Added as you can see below.

tests/conftest.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+@pytest.fixture(scope='session')
+def app(request):
+    """
+    This is creates the mock app for testing, it uses the
+    TestingConfig in Config.py
+    """
+    app = create_app(config_settings=TestingConfig)
+    ctx = app.test_request_context()
```
shots47s: Added the suggestion, after reading it is actually the right thing to do.  @liamocn give it another try.

Issue #76, opened 2019-08-29
Dynamic updating of datasets from the conp-datasets repo.


shots47s:
## Purpose
Currently, the datasets in our portal are statically updated at install time through a kludgy migration.  This should be dynamically populated through calls to whatever backend API that we are using, or pull from a registry process.

## Context
The portal should be dynamic in how it presents its datasets.

## Possible Implementation (optional)
1. Have this automatically done from time to time by polling the conp-dataset repo, potentially using datalad itself.
2. Create a means of registering datasets in the platform, and defining where they may be.

We could support both, but this is an open issue that would be good to discuss amongst the team and TSC before we move forward.

tomgee:
If all of the datasets referred to by the portal are in the conp-dataset repo, then an automatic poll would make the most sense.  However, if there are (or will be) datasets who do not reside directly in the repo, then presumably any changes to those externally-hosted datasets would involve pushing notification of the changes into the CONP portal/repo through a versioned registration process.

I think I'm not completely clear on the relationship between the portal and the repo, so perhaps my thought isn't quite worth the $0.02 I was hoping it was?

jbpoline:
All datasets should be in the conp-pcno/conp-dataset - pointing (as submodules) to their original datasets - ie, repositories. Some but not all of these repositories are living - names are super confusing - on the github conpdatasets organization. As long as we have access to the datasets from conp-pcno/conp-dataset then we should be able to populate the portal.

jbpoline:
Note that it would require
1- `clone conp-dataset`
2- `datalad install .`  for all datasets ...

shots47s:
Thanks @jbpoline, really all we need in the repository is the metadata, so I would qualify your previous point with that there needs to be at a minimum the dats.json with the appropriate information to populate the front page.  Also there needs to be a means to access the data where it lives, whether it be through datalad or some other means.  Right now, the portal is storing a url to the data, which is not a robust means to define how to access the data.  

But to populate the portal page with appropriate information, the metadata that describes the dataset should be in the conp-dataset repository.  

@jbpoline, have you and the team settled on a definitive list of things that are necessary in the dats.json?  I think so, but I thought it would be good to provide a pointer here in case others would like to comment.

jbpoline:
@shots47s  : yes, the dats files are compulsery, tests should fail if they dont exist !
yes - there is a list - need to dig out where (in Jen's documents) but here they are
* the schema.org required fields (title + description)
[...]

PR #90, opened 2019-09-13
Added dataset parser to cli.py

DESm1th:
---
name: Added a dataset cli parser
about: Added a crawler to make use of the dats model files
title: 'Feature implementation | Dataset Crawler'
labels: enhancement
assignees: ''

---

## Related issues

Related to: https://github.com/CONP-PCNO/conp-portal/issues/76

## Checklist

- [x] **DO** Unit tests pass.

## Purpose

This adds a 'update_datasets' cli option that will scrape the data sets in the portal's app/static/data/projects folder and either add a new record to the database (if the dataset ID and annex uuid are new) or find and update an existing record.

## Current behaviour

Currently the database 'datasets' table cannot be populated based on the datasets folder.

## New behaviour

The datasets table of the database can be populated based on the contents of app/static/data/projects by running `flask update_datasets`
 
- [ ] Yes
- [x] No

## Implementation Detail

There's a new function in app/cli.py. I added a few support functions as protected functions in that script because I wasnt sure if there was somewhere better to put them. Also, a few of the database columns I wasnt sure how to fill in: 'image', 'owner_id'  (users table not yet populated) and 'category'.

Also, (in spite of my best efforts) the code got a bit unruly. I think that it may be hard to adequately pull info out the dats.json files without normalizing a few of the columns in the current database schema. It might also make adding new datasets to the database easier if we stored the dats.json files as a 'jsonb' column in the database instead of trying to extract values, since we'd get the entire file contents if anything else in there ends up being useful later on,  but I think this might restrict the portal to using postgresql (I'm not sure which other rdbms support the json / jsonb column type).

I did not add any new unit tests but I can. I just wanted to get feedback to see if I went in the right direction first :)

### DIFF
--- a/app/cli.py
+++ b/app/cli.py
@@ -7,8 +7,15 @@ Module that contains the special command line tools
 import os
 import click
 import csv
+import configparser
+import json
+import logging
+from pathlib import Path
 from datetime import datetime
 from app.threads import UpdatePipelineData
+
+logging.basicConfig(format="%(name)s [%(levelname)s]: %(message)s")
+logger = logging.getLogger(os.path.basename(__file__))
 
 
 def register(app):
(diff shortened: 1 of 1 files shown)


app/cli.py
```diff
@@ -63,9 +70,146 @@ def seed_test_db():
...
+                name = None
+                modality = None
+                version = None
+                data_format = None
+                is_private = None
+
+            record = Dataset(dataset_id=dataset_id,
+                             annex_uuid=annex_uuid,
+                             description=description,
+                             owner_id=None,
```
shots47s: So owner is not in the DATS? right?

app/cli.py
```diff
@@ -63,9 +70,146 @@ def seed_test_db():
...
+                             download_path=download_path,
+                             raw_data_url=raw_data_url,
+                             image=None,
+                             name=name,
+                             modality=modality,
+                             version=version,
+                             format=data_format,
+                             category=None,
```
shots47s: Same with category, maybe we should consider changing this to "Keywords"

app/cli.py
```diff
@@ -7,9 +7,16 @@
 import os
 import click
 import csv
+import configparser
+import json
+import logging
```
shots47s: Adding logging.... I like it!

app/cli.py
DESm1th: There's a field that looks right, but the current database schema expects an integer so I assumed it was meant to link to an entry in the user table (there's no foreign key constraint on that column though so I could be wrong)

app/cli.py
```diff
@@ -63,9 +70,146 @@ def seed_test_db():
...
 
+    @app.cli.command("update_datasets")
+    def update_datasets():
+        from app import db
+        from app.models import Dataset
+        from sqlalchemy.exc import IntegrityError
+
+        logger.setLevel(app.logger.getEffectiveLevel())
+
+        data_path = app.config.get("DATA_PATH", "static/data/projects").lstrip("/")
```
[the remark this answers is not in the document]
shots47s: Actually, we should make this so it relies on the env variable more and defines only the relative path within the app.  So I would change this to something like:

`search_path = Path(app.config.get("DATA_PATH")).joinpath("projects")`

app/cli.py
```diff
@@ -63,9 +70,146 @@ def seed_test_db():
...
+                private = model.get("privacy") == "Open Access"
+            else:
+                description = None
+                download_path = None
+                raw_data_url = None
+                name = None
+                modality = None
+                version = None
+                data_format = None
+                is_private = None
```
shots47s: is_private is not defined outside of here, so if it doesn't hit this else, it doesn't know it.

shots47s:
Should make it so that if the data is not present for the dataset, no data is inserted into the database.  Right now, if a dataset fails to read a dats.json file, it gets an empty entry in the database with the annex ID.  Check to see if dats is there before committing to the database to keep the database pure.

app/cli.py
```diff
@@ -63,9 +70,146 @@ def seed_test_db():
...
+        value = parser[section][key]
+    except Exception as e:
+        logger.info("Couldnt parse {}.{} from {}. Reason - {} {}".format(
+                section, key, file_path, type(e).__name__, e))
+        return None
+    return value
+
+
+def _get_dats_model(project):
+    dats_file = project.joinpath("DATS.json")
```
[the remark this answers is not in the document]
xlecours: FYI https://github.com/conpdatasets/1000GenomesProject/pull/1
One offender fixed. Any suggestion on how to assert that the dats.json exists and is lowercase?

liamocn:
I'm inclined to go ahead and merge this if nobody's got any problems? There's a conflict that needs resolving but I don't think it's anything major.

shots47s:
I would hold off, what Xavier is working on will supersede this.

liamocn:
> 
> 
> I would hold off, what Xavier is working on will supersede this.

OK, will do

xlecours:
Following enhancement on the DATS.json standardization, a DATSDataset class was created to access the descriptor properties that can be in various formats.
Most of the 'ambiguous' columns have been removed from the Dataset table. 
The datasets are now installed and updated in app/cli.py as was proposed in this PR. But it is now using datalad to install the conp-dataset and subdatasets recursively. 

Closing this.

Issue #89, opened 2019-09-12
Improve test coverage in pytest


liamocn:
## Purpose
Having added the framework for testing the project, tests must now be added to increase overall test coverage

## Context
Test coverage can be monitored using coveralls, which has been added as part of #56

## Related issues (optional)
#56

cmadjar:
@liamocn any update on that issue?

liamocn:
@cmadjar We added a bunch of tests around the time this issue was opened, but the nature of testing is that you can always use more. I'd leave this open for now and I can work on some of the areas where testing could be improved when I get the time.

cmadjar:
Discussed at the CONP dev call of September 30th, 2020.

@liamocn will give a conceptualization demo of the test to be implemented for the portal at the next CONP dev call (October 7th)

PR #94, opened 2019-09-27
Loc #89

liamocn:
---
name: Increase test coverage
about: Adding tests to cover more endpoints
title: 'Feature implementation #89'
labels: enhancement
assignees: @liamocn 

---

## Related issues
#89 

## Checklist

- [x] **DO** Unit tests pass.
- [x] **DO** If new feature, created unit test.

## Purpose
Tests have been added to examine responses for search and pipeline endpoints

## Current behaviour
No tests for search and pipeline endpoints

## New behaviour
Now we create dummy data in the test database to test against. Requests made to dataset-search and pipeline endpoints check to ensure response is correct
 
- [ ] Yes
- [x] No

## Implementation Detail
Because of ongoing work on the portal's authentication, we still need tests to ensure responses are appropriate to whether the user is logged in, and what kind of access they should have

### DIFF
--- a/app/admin/routes.py
+++ b/app/admin/routes.py
@@ -14,66 +14,11 @@ from flask import render_template, request, flash, session, redirect, url_for,\
 from flask_login import current_user, login_user, logout_user, login_required
 from app.admin import admin_bp
 
-
-@admin_bp.route('/register_new_user', methods=['GET', 'POST'])
-def register_new_user():
-    """ Register New User Route
-
-        This currently registers a new user.
-        TO DO: Not entirely sure this is ever called and or works
-               Should redo with Flask-User
-
-        Args:
-            None
-
-        Returns:
-            if GET: the rendered register.html page
-            if POST: Creates new user and returns the rendered template
-    """
-    error = None
-    form = SignUpForm()
-
-    # Handle the BooleanField to check if is_pi or not
-    is_pi = False
-    if 'pi' in request.form:
-        is_pi = True
-
-    if request.method == 'POST':
-        # Check if passwords match
-        if request.form['password1'] == request.form['password2']:
-            if form.validate_on_submit():
-                # Check if email already exists
-                user = User.query.filter_by(email=request.form['email']).first()
-                # Create new user
-                if not user:
-                    user = User(
-                        oauth_id=request.form['orcid'],
-                        username=request.form['username'],
-                        email=request.form['email'],
-                        is_whitelisted=False,
-                        is_pi=is_pi,
-                        affiliation=request.form['affiliation'],
-                        expiration=datetime.now() + timedelta(6 * 365 / 12),  # 6 months
-                        date_created=datetime.now(),
-                        date_updated=datetime.now()
-                    )
-                    user.set_password(request.form['password1'])
-                    db.session.add(user)
-                    db.session.commit()
-                    login_user(user)
-                    return redirect(url_for('main.index'))
-            error = form.errors
-        error = {'Passwords do not match'}
-    return render_template('register.html', title='CONP | Register', form=form, error=error)
-
-
 @admin_bp.route('/admin')
 def admin():
     """ Admin Route
 
         Route that leads to the admin page
-        CURRENTLY NOT WORKING
-        TO DO: This needs to be checked against an admin role
 
         Args:
             None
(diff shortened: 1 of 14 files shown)


tests/conftest.py
```diff
@@ -18,13 +22,7 @@ def app(request):
     TestingConfig in Config.py
     """
     app = create_app(config_settings=TestingConfig)
-    ctx = app.test_request_context()
```
shots47s: Why do we not need the app context to be specifically specified?

app/admin/routes.py
```diff
@@ -14,66 +14,11 @@
...
-    return render_template('register.html', title='CONP | Register', form=form, error=error)
-
-
 @admin_bp.route('/admin')
 def admin():
     """ Admin Route
 
         Route that leads to the admin page
-        CURRENTLY NOT WORKING
-        TO DO: This needs to be checked against an admin role
```
shots47s: well, still really isn't an admin page :) . That is what I meant.  But this is fine.

tests/blueprint_specific_tests/profile_bp_tests/test_profile_endpoints.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+def test_pipelines_route(test_client):
+    """
+    GIVEN calling the route "/profile"
+    WHEN no user is logged in
+    THEN should return success code
+    """
+    res = test_client.get("/profile")
+    assert res.status_code == 200
```
shots47s: stray space at the end of file.

tests/blueprint_specific_tests/search_bp_tests/test_search_endpoints.py
```diff
@@ -0,0 +1,122 @@
...
+    session.add(new_dataset)
+    session.add(new_dataset_stats)
+    session.commit()
+
+    query = {'id': '8de99b0e-5f94-11e9-9e05-52545e9add8e'}
+    headers = {'Content-Type': 'application/json'}
+
+    # res = test_client.get("/dataset", headers = headers, query_string = query)
+
+    assert True
```
shots47s: stray space and need blank line at end.

tests/conftest.py
```diff
@@ -79,3 +77,80 @@ def test_client(app):
...
+    """
+    user = User(
+                oauth_id="0000-0000-0000-0000",
+                username="testuser",
+                email="example@mailinator.com",
+                affiliation="CONP",
+                expiration=datetime.utcnow() + timedelta(days=30)
+            )
+    user.set_password("ThisPassword")
+    return user
```
shots47s: extraneous space and need blank line at end.

tests/conftest.py
liamocn: I'm not sure. Sometimes the test run fails because the wrong context is 'popped' on teardown, but it's intermittent. I noticed that removing this line altogether solved the problem for me, but it's something to keep an eye on.

app/admin/routes.py
liamocn: There's no implementation at all as far as making this an admin page. I was just thinking about how we keep track of QA issues going forward

tests/conftest.py
shots47s: Ok, we can revisit, thanks.

tests/blueprint_specific_tests/admin_bp_tests/test_admin_endpoints.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+def test_admin_route(test_client):
+    """
+    GIVEN calling the route "/admin"
+    WHEN no user is logged in
+    THEN should return success code
+    """
+    res = test_client.get("/admin")
+    assert res.status_code == 200
```
driusan: shouldn't `/admin` return a 401 when no user is logged in?

tests/blueprint_specific_tests/search_bp_tests/test_search_endpoints.py
```diff
@@ -0,0 +1,123 @@
...
+"""
+import pytest
+from urllib.parse import urlparse
+from flask import url_for
+from flask_login import current_user, login_user, logout_user
+
+
+def test_pipelines_route(test_client):
+    """
+    GIVEN calling the route "/search"
```
driusan: it's a little confusing that `/search` is the endpoint but the test is named "test_pipelines_route" here..

liamocn: Ah, ok. That's a typo which I missed.

tests/blueprint_specific_tests/admin_bp_tests/test_admin_endpoints.py
liamocn: Yes it should. I'll update the test but there's also some implementation needed in general for the admin functionality

PR #99, opened 2019-10-07
Incorporating datasetstats in dataset

### LINKED ISSUE #97: DatasetStats needs to be linked one-to-one to a Dataset model and dynamically update as actions happen in the portal
## Purpose
Right now, there is no tie of the Stats to the Datasets, and they are not dynamically updated at all in the portal. 

## Context
The stats should be updatable and validated to actions in the portal.

## Possible Implementation (optional)
We should tie the stats to the Dataset via a one-to-one relationship and add hooks in the portal to updating when actions occur.

xlecours:
name: Add dataset stats in dataset table.
Fixes #97


xlecours:
@shots47s Please tell me on which branch I should send this.

PR #103, opened 2019-10-28
Fix os.path.expanduser call

driusan:
The update_pipeline_data cli command was calling expanduser with
an argument of '-' instead of '~' to represent the user's home
directory.

Fix the call to use the correct tilde expansion.


xlecours:
This is failing Travis.
And the issue seems to be fixed already

PR #104, opened 2019-10-28
Fix URL in pipeline search page

driusan:
The pipeline search page was pointing to the dataset-search URL.
Since both endpoints JSON have the same structure, this was resulting
in the page loading a list of datasets instead of a list of pipelines.

Fixing the URL is all that's required to fix the page since the JSON
of both endpoints is structured the same way.


driusan:
(Given that this is passing Travis, I think all of the other open PRs to this repo are failing because of this bug..)

Issue #95, opened 2019-10-01
Should updating pipelines add data to the database or should we rely on the file based search that is there now?


shots47s:
## Purpose
The updating of pipelines data from Zenodo happens at add places in the portal workflow by spawning threads.  These leads to inconsistent datasets, and the file based nature of this seems a bit fragile.

## Context
Should we have all of the data for the portal translated to the database, so that we have models to manipulate or is it best to rely on the cache files that are occurring now?  The database may be easier to index with something like elastic search.

## Possible Implementation (optional)
rewrite the system so that updating the pipeline data modifies the records in the database rather than being created on the fly from files in a .boutiques directory.

glatard:
`bosh` pulls descriptors in `.boutiques` and this won't be changed. The portal-specific file could be added to the database indeed, if this file becomes an issue. There shouldn't be any concurrent access to the file though.

driusan:
PR#105 and #103 fix the "spawning threads in odd places" issue and #104 fixes a (much more important) bug where the pipelines page was showing datasets instead of pipelines.

I'm not convinced that it's worth putting the data in the database instead of sharing the `~/.cache/boutiques` directory with bosh. There's currently an (unused) Pipeline database model defined in `app/models.py`, but the format of it doesn't match the bosh descriptors and I'm not convinced the file I/O is a significant bottleneck.

PR #105, opened 2019-10-28
Remove pipeline-search side-effects

driusan:
GitHub Issue #95 points out that pipeline cache updating
is currently done in odd places. In particular, it's done
by a background thread spawned by the pipeline-search endpoint
if either:

1. The cache directory doesn't exist
2. The cache directory is more than 5 minutes old

PR #103 fixes the bug which was preventing the
`flask update_pipeline_data` command from populating the cache,
which means the updating can be done on a periodic-basis from
cron now instead of spawning threads with side-effects when
searching for pipelines.

### DIFF
--- a/app/pipelines/routes.py
+++ b/app/pipelines/routes.py
@@ -61,28 +61,12 @@ def pipeline_search():
     all_desc_path = os.path.join(cache_dir, "all_descriptors.json")
     all_detailed_desc_path = os.path.join(cache_dir, "detailed_all_descriptors.json")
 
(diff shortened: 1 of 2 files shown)


xlecours:
@driusan Can you rebase this now that the tests are passing on master?

driusan:
@xlecours okay, done

xlecours:
@driusan failing travis; because pipelines aren't updated when gitting https://portal.conp.ca/pipelines

driusan:
@xlecours the tests are fixed

Issue #108, opened 2019-11-13
next page on the dataset page leads to page undefined


jbpoline:
## Current Behavior
I am getting :
https://portal.conp.ca/undefined?page=NaN&max_per_page=undefined
when I click on the arrow to go to the next page

shots47s:
Yeah, actually, pagination on DataTables React component needs to be completely redone, it currently doesn't work.

PR #110, opened 2019-11-26
FIx pagination issue

liamocn:
---
name: Fix for pagination issue #108 
about: This fix uses `setQuery` to update query state on navigation through the pages, instead of using hrefs.
title: Bug fix for issue #108'
labels: bug
assignees: ''

---

## Related issues
#108 

## Checklist

- [x] **DO** Unit tests pass.

- [ ] Yes
- [x] No

## Implementation Detail

 This fix uses `setQuery` to update query state when the user uses the page buttons. This updates the page attribute of the query state object that the `DataTable` is holding, and navigates through the pages, instead of using hrefs.


app/static/lib/conp-react/umd/1.conp-react.js
```diff
@@ -4,7 +4,7 @@
  */
 (window["webpackJsonpCONPReact"] = window["webpackJsonpCONPReact"] || []).push([[1],{
 
-/***/ 26:
+/***/ 29:
```
shots47s: Will these local changes have any substantial effect?  I guess this is because we are including the compiled react library in the repo.  Maybe we should consider making compiling the react components a part of the installation procedure.

PR #144, opened 2019-12-19
Remove invalid sort keys from datasets page and fix sorting with empt…

driusan:
…y values

Fix error where portal would throw an error if one of the fields had
`None` in a value. Also remove sort options which are meaningless from
list of options.

### DIFF
--- a/app/search/routes.py
+++ b/app/search/routes.py
@@ -109,12 +109,8 @@ def dataset_search():
             "title": d.name.replace("'", ""),
             "isPrivate": d.is_private,
             "thumbnailURL": "/dataset_logo?id={}".format(d.dataset_id),
-            "imagePath": "?",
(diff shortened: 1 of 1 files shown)


app/search/routes.py
```diff
@@ -131,7 +127,7 @@ def dataset_search():
     limit = max(min(int(request.args.get('limit') or 10), 10), 0)
     sort_key = request.args.get('sortKey') or "id"
     paginated = elements[(cursor):(cursor + limit)]
-    paginated.sort(key=lambda o: o[sort_key])
+    paginated.sort(key=lambda o: (o[sort_key] is None, o[sort_key]))
```
driusan: See https://stackoverflow.com/questions/18411560/python-sort-list-with-none-at-the-end

PR #147, opened 2019-12-19
white listing CONP datasets

xlecours:
This prevent the protal from displaying datasets that aren't on a white list. Meaning that only datasets in which CONP contribute to it's openness are to be displayed.

### DIFF
--- a/app/search/routes.py
+++ b/app/search/routes.py
@@ -100,8 +100,20 @@ def dataset_search():
     # Element input for payload
     elements = []
 
+    # CONP datasets whitelist
+    whitelist = [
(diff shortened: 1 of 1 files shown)


xlecours:
Note that this is a temporary measure. The conp-dataset repo will need to be "prune" for a permanent solution.

glatard:
Why don't we just clean up the conp-dataset? It would be way easier than having to keep and fix this white list. In particular, adding a dataset through the Zenodo crawler now requires to edit the white list to have it show in the portal.

xlecours:
The whitelist is now removed. 
Country indicator in place https://github.com/CONP-PCNO/conp-portal/pull/160

PR #148, opened 2019-12-20
Change spotlight dataset / other tweaks

liamocn:
---
name: Change spotlight dataset / other tweaks
about: Propose modifications in the code
labels: enhancement
assignees: @liamocn 

---

## Related issues

#143 
#128 
#145 

## Checklist

- [x] **DO** Unit tests pass.
- [ ] **DO** If new feature, created unit test.

#### Does this introduce a major change?
- [ ] Yes
- [x] No

PR #149, opened 2019-12-20
Update charts on dashboard

driusan:
This updates the chart legends s on the dashboard. It removes
the words "CONP" from the legend, since the chart contains
all datasets/pipelines, not only ones from CONP. It updates the
title to explicitly say that the bars are cumulative.

It also explicitly adds the Highcharts dependency to npm, since
otherwise the React app does not build.


driusan:
These changes are fixed by #148

Issue #153, opened 2020-01-15
Create a table of dataset with status


PapillonMcGill:
have 3 status:
CONP funded
Canadian dataset
External dataset (not canadian)

To be done by Naser

glatard:
Not sure how this table will be used, but if the idea is to represent these datasets differently in the portal, it might be better to add this information to the DATS model, and have the portal parse it.

PapillonMcGill:
It is a request by @samirdas to promote CONP funded datasets.

xlecours:
- DATS.json now contains the information as per https://github.com/CONP-PCNO/conp-dataset/pull/149
- @liamocn is working on using that information on the portal by displaying a canadian flag for   `DATS.json::extraProperties[@category="CONP_status"].values.value` is `CONP` or `Canadian` (missing value will default to `external` which won't display the canadian flag. ) PR to be sent shortly

PR #160, opened 2020-01-23
Add country indicator for datasets #153

liamocn:
---
name: Implementing the changes to the DATS model in the frontend
about: We can now use the 'CONPstatus' attribute to determine the country of origin of datasets
title: 'Feature implementation #153 '
labels: enhancement
assignees: ''

---

## Checklist

- [ ] **DO** Unit tests pass.

- [ ] Yes
- [x] No


app/search/models.py
```diff
@@ -75,6 +75,17 @@ def contacts(self):
...
 
+    @property
+    def conpStatus(self):
+        conpStatus = None
+        extraprops = self.descriptor.get('extraProperties', {})
+        for prop in extraprops:
+            if prop.get('category') == 'CONP_status':
+                conpStatus = prop.get('values')[0].get('value')
+
+        return conpStatus
```
glatard: There should be a default value for `conpStatus` or the portal will crash when the DATS model doesn't have this property. Suggestion: `external`.

liamocn: Thanks for the feedback, I've implemented that now.

app/templates/fragments/navigation_bar.html
```diff
@@ -42,8 +42,8 @@
             >
               Help
             </a>
-            <div class="dropdown-menu" aria-labelledby="navbarDropdown">
-              <a class="dropdown-item" href="/#/">Documentation</a>
+            <div class="dropdown-menu dropdown-menu-right" aria-labelledby="navbarDropdown">
+              <a class="dropdown-item" href="https://github.com/CONP-PCNO/conp-portal">Documentation</a>
```
xlecours: https://github.com/CONP-PCNO/conp-portal/wiki
```suggestion
              <a class="dropdown-item" href="https://github.com/CONP-PCNO/conp-portal/wiki">Documentation</a>
```

Issue #182, opened 2020-02-29
Show dataset authentication status in the portal


glatard:
## Purpose

DATS models have recently been extended with an authorization field, that can be "public", "registered" or "private". It would be useful to represent that in the portal, with some kind of icon that could be next to the Canadian flag.

## Context

This would allow users to quickly identify which datasets can be downloaded without authentication.

liamocn:
Can you give a better idea of the difference between 'registered' and 'private', and what type of icons you think might be suitable? I think using padlocks or security type icons might be a little heavy for an open science project.

glatard:
* private means that you need to negotiate custom credentials with the platform hosting the data
* registered means that the dataset is accessible using a CONP portal login (not available yet)

PR #190, opened 2020-03-12
Show authentication on datasets

liamocn:
---
name: Show authentication on datasets
about: Modify the dataset view to show what accounts are required for dataset download
title: 'Feature implementation #182 '
labels: enhancement
assignees: ''

---

## Related issues

#182 

## Checklist

- [x] **DO** Unit tests pass.

- [ ] Yes
- [x] No

## Implementation Detail

Icons are used a the bottom of the dataset view to show where accounts are required in order to download a dataset. If one is required, it can be either a CONP account, or a third-party account at the dataset source.

NOTE: Accounts on the CONP portal were removed in December and have not been reinstated. We also lack details on how we should direct users in the case of third-party accounts being required. As a result, this pull request may be confusing if merged now.


xlecours:
@liamocn Can you rebase this PR? We merge a change in requirement.txt that updates MarkupSafe to 1.1.1; which solves the issue in Travis with pip install.

liamocn:
Sure, will do that now.

PR #195, opened 2020-03-18
Add new property to DATSDataset to generate schema.org json-ld

zxenia:
---
name: New property to DATSDataset to generate schema.org json-ld
labels: enhancement

---
## Checklist

- [x] **DO** Unit tests pass.
- [x] **DO** If new feature, created unit test.

## Purpose
Generate a schema.org json-ld snippet for Google dataset search from a valid DATS.json.
The produced json-ld should be embedded in html so that google can extract the metadata and make a dataset discoverable.

## New behavior
DATSDataset class has a new property `schema_org_metadata` that returns schema.org json-ld.
 
- [ ] Yes
- [x] No

## Implementation Detail
Added a new property to existing class DATSDataset because it makes use of `self.descriptor`.
Added unit tests here `tests/unit_tests/test_datsdataset_schema_org.py`

The schema.org json-ld object can be tested in [Google Rich Results Test](https://search.google.com/test/rich-results)

The schema.org json-ld should be added here, e.g.:
```
<html>
  <head>
    <title>Test dataset</title>
    <script type="application/ld+json">
    {
      "@context":"https://schema.org/",
      "@type":"Dataset",
      ......
    }
    </script>
  </head>
  <body>
  </body>
</html>
```

Also updated MarkupSafe to a newer version 1.1.1 due to this issue https://github.com/pallets/markupsafe/issues/57

### DIFF
--- a/app/search/models.py
+++ b/app/search/models.py
@@ -205,7 +205,6 @@ class DATSDataset(object):
         return "{}".format(sources)
 
 
-
     @property
     def subjectCount(self):
         count = 0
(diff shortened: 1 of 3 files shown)


zxenia:
Hi @liamocn 

Here is how the schema.org json-ld should be added to the front-end. I don't know the conp-portal codebase well so this solution is approximate.

Class DATSDataset (app\search\models.py) has a property `schema_org_metadata` which returns json-ld required by Google Dataset search.

1. In `app/search/routes.py` https://github.com/CONP-PCNO/conp-portal/blob/master/app/search/routes.py#L319

function `dataset_info()` should also include `schema_org_metadata` value (or another option would be adding to `get_dataset_metadata_information()`)
e.g.
```
dataset = {
    "schema_org_metadata": datsdataset.schema_org_metadata,       # add this
    "authorized": authorized,
    "name": datsdataset.name,
    ... etc.
}
```

2. In `app/templates/dataset.html`

I think it can be added to the block scripts since it has to be placed inside of the `<head>` tag:

`<script type="application/ld+json">{{ data.schema_org_metadata }}</script>`


When we go to an individual dataset webpage we should see in the `<head>` something like e.g.:

```
<head>
    <script type="application/ld+json">
    {
      "@context":"https://schema.org/",
      "@type":"Dataset",
      "name": "Braincode....",
      ......
    }
    </script>
</head>
```

After it's in the portal it might take a few days for Google to crawl the metadata in order to be added to the Google Dataset search.

Issue #158, opened 2020-01-22
Write the term of use content to display


xlecours:
This issue is to generate the term of use to display to the user upon first visit.

@PapillonMcGill I think this should be done with Naser

glatard:
Waiting for the terms of use.

jbpoline:
See PR #194

xlecours:
> See PR #194

@jbpoline #194 is not a PR but an issue

xlecours:
Got it, #192 is the PR

xlecours:
This is resolved.

PR #192, opened 2020-03-18
Adding a first term of use for the portal

jbpoline:
The term of use should be refined by the ethics and governance team

---
name: Term of use file addition
about: Term of use file addition
title: 'Term of use file addition'
labels: enhancement
assignees: ''
---

## Checklist

- [ ] **DO** Unit tests pass.
- [ ] **DO** If new feature, created unit test.

#### Does this introduce a major change?
- [ ] Yes
- [x] No

PR #203, opened 2020-03-24
Adding content in template

xlecours:
## Related issues
https://github.com/CONP-PCNO/conp-portal/issues/158

## Current behaviour
"term of use goes here" is displayed

## New behaviour
actual content is displayed
 
- [ ] Yes
- [x] No

## Implementation Detail
The template file contains the text. It might be desirable to have the content in a dedicated file.

### DIFF
--- a/app/templates/common/base_main.html
+++ b/app/templates/common/base_main.html
@@ -54,7 +54,66 @@ CONP Portal
         </button>
       </div>
       <div class="modal-body">
-        Terms go here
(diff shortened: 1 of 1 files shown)


xlecours:
@liamocn can take this and make it "pretty"?

Currently, the modal width limit the display and since it was provided as markedown in https://github.com/CONP-PCNO/conp-portal/pull/192 the text is somewhat plain...

Issue #205, opened 2020-03-25
Remove sorting capabilities for Modality, Format and Source


cmadjar:
## Expected Behavior

Remove sorting capabilities for Modality, Format and Source. Instead add those DATS fields to the search engine so one can filter for datasets containing MINC etc...

## Context (optional)

Discussed during the CONP Dev call of March 25th.

cmadjar:
Links to the following issue regarding adding DATS fields in the search https://github.com/CONP-PCNO/conp-portal/issues/183

emmetaobrien:
I am at this point sure the DATS fields are consistently structured for finding this data, but there may well be places where the content is not what someone familiar with the dataset would expect, e.g. Naser mentioned SIMON and the recently generated derived BigBrain not having MRI listed as a modality.  It is straightforward for me to update any cases here; in the case of SIMON I believe @glatard is going to recrawl it soon and it would probably be better practice to generate this data at that point than for me to manually edit it.  Let me whatI should modify here.

Issue #200, opened 2020-03-24
Sorting tools and pipelines by title should be case insensitive


cmadjar:
## Expected Behavior

When sorting tools and pipelines by title, one would expect that it to not be case sensitive. 

## Current Behavior

After sorting tools and pipelines by title, tools and pipelines starting with a capital letter are listed first and follow several pages later by tools and pipelines starting with a lower case letter.

## Context (optional)

This issue was reported by Naser when he reviewed the portal before release.

cmadjar:
Just checked the portal with the latest code pulled. The issue is fixed. Thank you @liamocn !!

PR #215, opened 2020-03-30
Update dataset sorting; add filters for modality, format

liamocn:
## Related issues

#200 #205 

## Checklist

- [ ] **DO** Unit tests pass.

## Purpose

Fix case-sensitivity when sorting by title. Use separate sorting logic for different sort types.

Remove sorting by modality, format and source from the dataset search page. Instead, add filters for modality and format which can be checked on or off to search by those modalities and filters.
 
- [ ] Yes
- [ ] No


app/search/routes.py
```diff
@@ -137,18 +138,43 @@ def dataset_search():
...
-        paginated = paginated[(cursor):(cursor + limit)]
+
+        if(sort_key == "conpStatus"):
+            order = {'conp': 0, 'canadian': 1, 'external': 2}
+            paginated.sort(key=lambda o: order[o[sort_key].lower()])
+
+        elif(sort_key == "title"):
+            paginated.sort(key=lambda o: o[sort_key].lower())
+
+        elif(sort_key == "size"):
```
xlecours: This is not working correctly. It should support missing values (as in the final else) and 1 GB shows as smaller than 194 MB.
Sorting by size could be done in subsequent PR since it is not trivial (unit convertion).

liamocn: I was going to look at sizes as a separate PR, but it might make more sense to do it here. Let me take a look.

app/search/routes.py
```diff
@@ -137,18 +138,52 @@ def dataset_search():
...
+
+            def getAbsoluteSize(e):
+                units = ["KB", "MB", "GB", "TB"]
```
xlecours, line 168:
```suggestion
                if not e["size"]:
                    return 0.0

                units = ["KB", "MB", "GB", "TB"]
```

app/search/routes.py
```diff
@@ -137,18 +138,52 @@ def dataset_search():
...
+                unitScales = [1000, 1000**2, 1000**3, 1000**4]
+                size = e["size"].split(" ")
+                absoluteSize = size[0]
+                if(size[1] in units):
+                    absoluteSize = float(size[0]) * unitScales[units.index(size[1])]
+                return absoluteSize
+
+            paginated.sort(key=lambda o: (o[sort_key] is None, getAbsoluteSize(o)), reverse=True)
```
xlecours, line 176:
```suggestion
            paginated.sort(key=lambda o: getAbsoluteSize(o), reverse=True)
```

cmadjar:
@liamocn There is a conflict on this PR probably due to the changes of PR #226 that just got merged. Sorry...

Issue #208, opened 2020-03-25
Disable link for greyed out CBrain buttons


cmadjar:
## Purpose

When a user clicks on a grey CBrain button, they are getting redirected. Instead, we should disable the ability to click on a grey CBrain button to avoid any confusion.

## Context

Discussed during the CONP dev call of March 25th, 2020.

PR #219, opened 2020-03-31
Disable grey cbrain

liamocn:
## Related issues

#208 

## Checklist

- [x] **DO** Unit tests pass.

## Purpose

When shown they grey CBrain link in the pipeline search page, the link is now disabled.

- [ ] Yes
- [x] No


cmadjar:
@glatard @xlecours When you get a chance if you could review this PR that would be great. I looked at it and it seemed all good but I don't have your experienced eyes :).

xlecours:
I think there should be no grey icon; only a blue icon when the tool is available on cbrain.
The text should be "available on CBRAIN" since this is not linking to a page to launch the tool but only to login to CBRIAN. 

Also, the icon is now really bif on my screen. Can we put it to it's origin size?

liamocn:
> 
> 
> I think there should be no grey icon; only a blue icon when the tool is available on cbrain.
> The text should be "available on CBRAIN" since this is not linking to a page to launch the tool but only to login to CBRIAN.
> 
> Also, the icon is now really big on my screen. Can we put it to it's origin size?

I changed the way the images scale by screen size: that image may be too large now if given a lot of horizontal space. I can change this. Regarding removing the grey icon altogether, let's talk about that on our next call just so we're sure about it.

Issue #225, opened 2020-04-01
DataLad commands for openpain datasets incorrect


cmadjar:
The DataLad instructions are incorrect for all subdatasets of the openpain project.

Current installation information:
```
 ~$ datalad install https://github.com/CONP-PCNO/conp-dataset.git
 ~$ cd conp-dataset
 ~$ datalad install openpain-subacute_longitudinal_study
 ~$ cd openpain-subacute_longitudinal_study
 ~$ datalad get <filepath>   
```

But in reality, it should be:
```
 ~$ datalad install https://github.com/CONP-PCNO/conp-dataset.git
 ~$ cd conp-dataset
 ~$ datalad install openpain/subacute_longitudinal_study
 ~$ cd openpain/subacute_longitudinal_study
 ~$ datalad get <filepath>  
```

Basically, there should be a `/` between `openpain` and `subacute_longitudinal_study` in line 3 and 4 of the copy-pasted code above.

cmadjar:
Fixed by 226

PR #226, opened 2020-04-02
Fix truncated dataset names

### LINKED ISSUE #181: Download options truncates long names
Long dataset names are truncated in the Download Options box, which makes download commands invalid. For instance, for dataset `MRI_and_unbiased_averages_of_wild_muskrats__Ondatra_zibethicus__and_red_squirrels__Tamiasciurus_hudsonicus_`:

Screenshot from 2020-02-29 16-12-27

xlecours:
Fixes #181 

Also change the Download option presentation to make the commands copy-paste-able.

Before:
before

After:
after

### DIFF
--- a/app/templates/dataset.html
+++ b/app/templates/dataset.html
@@ -53,18 +53,18 @@
   </div>
 </div>
 
+<h2 class="card-title">Download Options</h2>
+Download is enabled through <a style="color: blue;" href="http://handbook.datalad.org/en/latest/intro/installation.html#install">datalad</a>
+<br>
+Here are commands to download content from this dataset:
 <div class="card">
   <div class="card-body">
-    <h2 class="card-title">Download Options</h2>
     <pre class="card-text" style="white-space: pre-wrap;">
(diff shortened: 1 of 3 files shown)


app/templates/dataset.html
```diff
@@ -53,18 +53,18 @@ <h2>CONP Portal | {{ data.title }} </h2>
...
-      
-# Here are commands to download content from this dataset:
-    ~$ datalad install https://github.com/CONP-PCNO/conp-dataset.git
-    ~$ cd conp-dataset
-    ~$ datalad install {{data.id}}
-    ~$ cd {{data.id}}
-    ~$ datalad get &ltfilepath&gt
+datalad install https://github.com/CONP-PCNO/conp-dataset.git
+cd conp-dataset
+datalad install projects/{{data.name}}
```
cmadjar, line 65: This looks good! The only thing maybe is that the name is only the basename of the directory. However, for projects like openpain subdatasets it should be `projects/openpain/basename` because that is where those are stored. 

Is there a way to include parent directory when there is one in the directive?

cmadjar: #225

xlecours: Nice catch!

Issue #238, opened 2020-04-08
Links should open in a new tab


cmadjar:
Links displayed in the portal should open in new tabs instead of the current page.

Exception being the buttons/links on the left side of the screen.

PR #244, opened 2020-04-08
open external links in new tab

liamocn:
## Related issues

#238 

## Checklist

- [ ] **DO** Unit tests pass.

## Purpose

This opens links to zenodo, datalad and dataset sources in a new tab

**NOTE**: This will not currently work with links within the share and FAQ pages, as Github generates the links and it's not the behaviour there.
 
- [ ] Yes
- [ ] No

## Implementation Detail

The <a> tags have been changed to open in a new tab where appropriate.


cmadjar:
@liamocn this is failing travis.

liamocn:
> 
> 
> @liamocn this is failing travis.

With the same boutiques issue you were having. It's not the code change that's causing the failure and I'm unsure how to fix it.

Issue #240, opened 2020-04-08
Links within the MD file does not seem to work in the displayed Share MD file


cmadjar:
The link at the top should redirect to the different sections of the MD file. Those links work when being clicked on on Github but does not seem to work on the Share page of the portal.

Confirm that it is the same issue with the FAQ MD files when it will be integrated in the portal and figure out how to fix this.

PR #256, opened 2020-04-13
Fix deep links, update bootstrap. Changes to main jinja template structure.

liamocn:
## Related issues

#240 

## Checklist

- [ ] **DO** Unit tests pass.

## Purpose

This PR fixes a problem with non-functional deep links on the Share and FAQ pages.
 
- [x] Yes
- [ ] No

## Implementation Detail

The Github markdown API creates name attributes in the style 'user-content-*' for all <a> tags in the returned html. This breaks deep-linking since the target attribute on the link does not contain 'user-content-'. Removing the prefix fixes the links.

However, the Bootstrap Navbar that we use has a quirk in that it overlays itself on the content underneath. We use padding to offset content on page load but this does not work for deep-linking since the top of the scroll view is still underneath the navbar. A further css workaround was required to fix this.

Bootstrap has been updated to the latest, and Flask-Bootstrap removed since it is 2 years out of date. `base_public.html` and `base.html` have been removed so that all templates inherit the same base template, `base_main.html`


liamocn:
Flask-Bootstrap was added back since it's need for some of the user account tempates.

Issue #260, opened 2020-04-15
Modality selection filter not working for Quality Control Subject and Basic Demographics


cmadjar:
This could be due to a spacing or case sensitive issue.

In the filter selection tab, they are written: `qualityControlSubject` and `basicDemographic` but in the data, they are written `Quality Control Subject` and `Basic Demographic`

Issue #253, opened 2020-04-09
Sort function: should have the ability to sort up and down


cmadjar:
I should have the ability to sort up or down. I probably want to look at the bigger datasets before looking at single subject niche datasets but now I have to scroll down. 

## Context (optional)

Alan's feedback

Issue #252, opened 2020-04-09
Sort function: explain what each criterion means


cmadjar:
For instance, Size might mean the number of subjects or number of files but it actually means total disk space. etc...

## Context (optional)

Alan's feedback

cmadjar:
In the datasets page, the following should be relabelled:

- origin -> ?, not sure what that means actually...
- title -> dataset name
- size -> disk space usage
- files -> number of files
- subjects -> number of subjects
- format -> file format

liamocn:
'Origin' refers to whether the datasets are canadian, external etc.

PR #263, opened 2020-04-16
Sort updates

liamocn:
## Related issues

#252 #253 #260 

## Checklist

- [ ] **DO** Unit tests pass.

## Purpose

Fixes sorting issues with dataset search. Sorting is now possible in ascending and descending orders.
 
- [ ] Yes
- [x] No


cmadjar:
@liamocn Another thing I just noticed. In the modality filter, histology is duplicated (histology vs Histology). There should be only one of them available in the filter and it should be able to filter for both cases.

Same thing for file format (minc/MINC should refer to the same thing)

liamocn:
> 
> 
> @liamocn thank you for the updates.
> 
> A few notes regarding the modality and file format filters:
> 
>     * Modality:
>       
>       * eeg should be displayed as EEG
>       * t1w should be displayed as T1W
>       * mri should be displayed as MRI
>       * fmri should be displayed as fMRI
> 
>     * File format
>       
>       * there is still duplication of file formats (csv vs .csv, nifti present twice and many others)
>       * also, I think it would be best to display the name of the format with their official way of writing them
>         
>         * nifti => NIfTI (note the lower case f in this one)
>         * minc => MINC
>         * dicom => DICOM
>         * tsv => TSV
>         * csv => CSV
>         * json => JSON
>         * and so on
> 

@cmadjar It's difficult to determine any of this on the frontend since there's a certain amount of variation in the datasets themselves. Ideally, keys for modality and format would be unique and consistent so they can be properly indexed. If they had a display name or similar associated with them, which would be the readable form, even better

cmadjar:
@liamocn Thank you for your reply. These can be dealt with later on. 

However, did you see my last comment about the selection of formats not being displayed the same way as the selection of modality is displayed? (a.k.a. it goes up the screen instead of being displayed below the filter)

image

liamocn:
@cmadjar Yes, I'll see if I can resolve that. Thanks.

Issue #245, opened 2020-04-08
Make the dataset download instructions more user-friendly


cmadjar:
While all the correct information is present in the dataset page, need to make the download instruction a bit more user-friendly:
- explicitly mention that in order to download the dataset, users need to have basic understanding of UNIX command lines
- explicitly mention that in order to download the dataset, users need to install DataLad
- explain to the user what the different commands do and what to expect from those commands
- display the README.md on the dataset page, those README would contain information that datasets people think important to understand what their dataset is about and how it is organized.

Will start drafting the page in that ticket later on and seek feedback before going into development phase. Assigning it to me for now.

cmadjar:
![Screen Shot 2020-04-08 at 4 38 51 PM](https://user-images.githubusercontent.com/1402456/78831341-74932d80-79b7-11ea-9f80-ed7132607ffd.png)


### Dataset Download Instructions

![Screen Shot 2020-04-08 at 4 39 37 PM](https://user-images.githubusercontent.com/1402456/78831395-912f6580-79b7-11ea-9de0-5a08f8244ddd.png)

The following instructions requires to have a basic understanding of UNIX command lines. We are currently working on a way to download the dataset via a button on the browser for easier download.

Dataset download is enabled through [DataLad](https://www.datalad.org/). To install DataLad on your system, please refer to the [install section of the DataLad Handbook.](http://handbook.datalad.org/en/latest/intro/installation.html#install). Installation via `miniconda` is recommended in order to obtain the latest version of DataLad. 

To initiate the CONP dataset (`conp-dataset`), run the following command in the directory where you want CONP datasets to be installed:
```
datalad install https://github.com/CONP-PCNO/conp-dataset.git
```

To install the dataset, go into the created `conp-dataset` directory and run `datalad install` on the dataset:
```
cd conp-dataset
datalad install projects/**DATASETNAME**
```

Now that the DataLad dataset has been installed, go into the dataset directory under `projects/**DATASETNAME**`. 
```
cd projects/**DATASETNAME**
```

Note that files present in the dataset directly are symlinks and will need to be downloaded manually using the `datalad get <filepath>` command in the dataset directory:
```
datalad get <filepath>
```
Note, if you run `datalad get *` command, all the files present in the dataset directory will be downloaded. 

For more information on how DataLad works, please visit the [DataLad Handbook documentation](http://handbook.datalad.org/en/latest/).


### Dataset README information

[...]

PR #268, opened 2020-04-17
use function instead of internal API call for readme content

liamocn:
## Related issues

#245 

## Checklist

- [ ] **DO** Unit tests pass.

## Purpose

This is intended to fix an issue where the readme content returned was incorrect
 
- [ ] Yes
- [x] No


cmadjar:
Pulled the code on production and the issue is fixed. Thank you @liamocn !

Issue #271, opened 2020-04-22
Add a contact us page that will display the google form 


cmadjar:
## Purpose

Create a contact us form using google form and display it on the portal.

Steps involved:

1) Create the Contact Us google form (ensure notification is turned on)

2) Create a "Contact Us" link next to the Terms of use button that will open a new window with the google form embedded in it

cmadjar:
The google form has been created in the [CONP-Cyberinfrastructure-TSC](https://drive.google.com/open?id=1_EQW8xltIQvKiGlrvtXxZqyUEXADvYUs) shared google drive:

https://drive.google.com/open?id=1Skamop7PZALK67-ckLpd9YweJLJ0FmhE 

- Full link to the form: https://drive.google.com/open?id=1H5uYNBZFxttGRuZwHfHQwT2NjdzYFTt0ojKgUXjs1h8

- Link to the answers: https://drive.google.com/open?id=1g0DD0L1XHj3vxPi0qHbp7uGL0eftU9Pus0g2UmXGczM

PR #273, opened 2020-04-27
Add contact us form

liamocn:
## Related issues

#271 

## Checklist

- [ ] **DO** Unit tests pass.

## Purpose

This adds a 'Contact Us' with a navigation link in the top bar. The content itself is and embedded google form
 
- [ ] Yes
- [ ] No

## Implementation Detail

Google forms allow you to embed a form using an iframe. This is simply added to the html in the page content

### DIFF
--- a/app/main/routes.py
+++ b/app/main/routes.py
@@ -87,3 +87,19 @@ def faq():
     content = response.text.replace('user-content-','')
 
     return render_template('faq.html', title='CONP | FAQ', user=current_user, content=content)
+
+
+@main_bp.route('/contact_us')
+def contact_us():
+    """ Contact Us Route
+
+        Route to lead to the contact page
+
+        Args:
+            None
+
+        Returns:
+            rendered template for contact_us.html
+    """
+
+    return render_template('contact_us.html', title='CONP | Contact Us', user=current_user)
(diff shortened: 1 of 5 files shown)


liamocn:
contactus
@cmadjar This is what I see. I think it's that you're left at the bottom of this page when you submit the form. Can you see if that's the case for you?

cmadjar:
Oh, I see. Is there a way to make it so the page is scrolled backed up after from submission? Actually, is there a reason why the page is scrollable when there is so much blank? Could this be fixed somehow? 

Thanks!

liamocn:
Yes, I'll investigate.

cmadjar:
@liamocn Just had a flash about something else. The Contact us at the end of the page should be removed as long as the mysterious info@conp.ca email address that does not seem to reach anyone.

Also, institutions@conp.ca does not exist so I would remove the entire foot Contact us section in favour of the form.

Thanks!

liamocn:
@cmadjar So we can't actually get any sense of what content is being displayed inside the iframe at any point (this is a feature of iframes), and since it isn't our iframe we can't implement any callbacks to get around this. I've resized the iframe now but I'm not sure the experience is better since there's an internal scrollbar, but you will see the response when you send the form. Let me know what you think.

Issue #270, opened 2020-04-22
Datasets update broken


xlecours:
There is a new dataset in conp-dataset with a long name. here is the error log of `flask update_datasets`

`
sqlalchemy.exc.DataError: (pymysql.err.DataError) (1406, "Data too long for column 'dataset_id' at row 1")`

glatard:
Are there any news about this? Two datasets were added recently and don't seem to show in the portal, this is a bit embarrassing.

cmadjar:
That is not good... Who could this be assigned to? I am not sure who is the best to tackle that one.

glatard:
Thanks @liamocn for taking this on!

liamocn:
@glatard @xlecours @cmadjar OK, there's a PR in for this now. I believe `flask db upgrade` will need to be ran on the production server which I don't have access to. Regarding the datasets which failed, it's very possible they will be considered 'installed' by the update_datasets script even though inserting the records failed, though I'm not 100% sure. In that case, removing the datasets directory entirely and cloning it fresh is recommended.

cmadjar:
@liamocn @xlecours I tried running `flask db upgrade` on the production server but got the following error:
```
conp-admin@portal:~/conp-portal$ flask run upgrade

Command 'flask' not found, but can be installed with:

apt install python3-flask
Please ask your administrator.
```

I am surprised flask is not installed on production. @xlecours Would you mind taking a look at production for that? J'ose pas, LOL. Can't wait to have a dev version of the portal so I would not feel too stressed about messing up ;).

liamocn:
@cmadjar It sounds like you didn't activate the python environment, with `source venv/bin/activate`

cmadjar:
ooops, right. Did not think of it... I did run `flask db upgrade`. Not sure what to expect from then though

glatard:
Hi, are there updates about this? The portal still doesn't seem to have the new datasets.

cmadjar:
@liamocn will look at what is going on on the portal.

Access to the portal.conp.ca VM has been asked for Liam: https://redmine.cbrain.mcgill.ca/issues/17864

liamocn:
I migrated the production db to latest revision (though there were some issues with the migration scripts) and updates the datasets:

`(venv) conp-admin@portal:~/conp-portal$ flask update_datasets
[...]

PR #274, opened 2020-04-29
Fix dataset id bug

liamocn:
## Related issues

#270 

## Checklist

- [ ] **DO** Unit tests pass.

## Purpose

This increases the size of the dataset_id in the dataset db model to accommodate large dataset names. Flask-Migrate has also been set to migrate schema changes of this sort.

**NOTE:** following this update `flask db upgrade` will need to be ran in the prod environment if it's not already part of the deployment script.

 
- [ ] Yes
- [x] No


glatard:
Hi @liamocn thanks for working on this! Just a question: what is going to happen in the unlikely event that a file name is longer than 255 chars? Maybe, just in case, directory names could be truncated to 255 chars before trying to add them to the DB.

liamocn:
@glatard Good point. The way it is now, the name and dataset_id are the same length. If we truncate then the records will be inserted ok, but there will be display issues in the portal since the records are used to populate the datasets in the portal. In fact we can make those fields as long as we like. The only reason SQLite cares is for portability with MySQL or Postgres etc.

Issue #242, opened 2020-04-08
Redesign the tools & pipelines page


cmadjar:
The Tools & Pipelines page should be redesigned to be similar to the way data are organized:

- the landing page would list the pipelines and tools and users could click on specific tools which will redirect to a detailed tool page. The summary information displayed in the pipeline row could be the following from the Boutiques Schema:

  - description, truncated to the first X character
  - name
  - tags
  - zenodo id / doi

- create a detailed too page that would read the Boutiques Schema and display at least those information:

  - author
  - container-image (en particulier le type, pour savoir si c'est Docker ou Singularity)
  - description
  - descriptor-url
  - doi / zenodo id
  - name
  - online-platform-urls (c'est ce qui est derrière le cbrain button)
  - tags
  - tool-doi
  - tool-version
  - tool-url

In addition, in the detailed page, we should add clear instructions on how to run the tools with also clear links to the Boutiques Tutorial

Link to other issues regarding the design of the tools & pipelines page:
- https://github.com/CONP-PCNO/conp-portal/issues/229
- https://github.com/CONP-PCNO/conp-portal/issues/224
- https://github.com/CONP-PCNO/conp-portal/issues/209
- https://github.com/CONP-PCNO/conp-portal/issues/208

glatard:
Link to the Boutiques schema doc: https://github.com/boutiques/boutiques/tree/master/schema

cmadjar:
Link to the sketch of the website that was done by Derek and Santiago a year ago. We could probably start from that and build on top of it. To be discussed in a meeting early next week.

https://sketch.cloud/s/mOnl8 

**a few notes and quick ideas:** 
- make the tools page like the data page (list of rows) with the information mentioned in the description of this issue (including tool logos when provided otherwise default wheels logo, icon grey/blue to run on CBRAIN, URL to the tool...) 
  Also ensure to include the number of rows in the data table so users know how many tools are available easily 
- in the detailed page
  - short instructions on how to run this locally (with links to Boutiques tutorial)
  - run on CBRAIN button
  - URL to the tool
  - Zenodo URL
  - detailed list mentioned on the description of this ticket
  - dated added, and date updated if that information is easily findable?

glatard:
I think the 1st page should contain the following info:
-     description, truncated to the first X character
-     name
-     tags
-     zenodo id / doi
-    cbrain button

Before we meet, I'd suggest that someone drafts a mock interface so that we have something concrete to discuss from.

[...]

PR #275, opened 2020-05-04
Pipelines page

liamocn:
## Related issues

#242 

## Checklist

- [ ] **DO** Unit tests pass.

## Purpose

This adds a pipeline details page with information and execution options for the user

Design here:
https://www.sketch.com/s/mOnl8/a/dq3g22
 
- [x] Yes
- [ ] No


liamocn:
@cmadjar OK, I've made most of those changes now.

> 
> Filter page:
> 
>     * Is there a way to make the links from descriptions show up as links? See example of the "BIDS app example" tool
> 

The problem with making links out of a description which may or may not contain a URL is that we must generate the html component at runtime just for incidences of those URLs where they crop up. React in particular does not like this and it's considered unsafe, so I'd recommend against it.

Issue #277, opened 2020-05-08
Pipeline page updates


liamocn:
1. "Run externally" should not point to the descriptor URL on Zenodo, because it's not possible to run a tool on Zenodo. Instead, it should link to the URL in online-platform-urls when this URL isn't a CBRAIN one.

2. Online platform URL shouldn't appear in the list below

![Screenshot from 2020-05-07 14-14-24](https://user-images.githubusercontent.com/8952928/81428916-03b47200-912b-11ea-8cf5-52df0ebf16dd.png)

3. When a property has no value, it should be omitted rather than shown as "N/A".

4. Container image should contain the name of the Docker or Singularity image, not just the image type.

5. Tags could be shown a bit more prominently, similar to issue labels in GitHub.

liamocn:
@glatard For clarity, how would I go about displaying the container information below? Are all of these fields to be included? What are the links, if any?

"container-image": {…}
​​image: "mcin/docker-fsl:latest"
​​index: "index.docker.io"
type: "docker"

glatard:
Hi Liam, the link to the container must be assembled from the index and image, and I'm not sure how to do that in a robust way. I would suggest to just show the container type, image and index and let the user deal with it. For Docker anyway, container images are most often retrieved using the Docker client which works with image names, not URLs.

PR #278, opened 2020-05-12
Pipeline updates

liamocn:
## Related issues

#277 

## Checklist

- [ ] **DO** Unit tests pass.

## Purpose

This addresses the issues specified in issue #277
 
- [ ] Yes
- [ ] No

## Implementation Detail

The "run externally" section is now reserved for non-CBrain platforms where available. I couldn't find any working examples from the current pipelines, however.

The 'container image' still just reports the image type as I'm unsure what to do with it

Fields without a value are omitted

Tags are now much more apparent in the pipeline view. They are also now clickable, and will search for pipelines by the respective tag,


cmadjar:
@liamocn It looks like the container that is supposed to link to CBRAIN does not work for tools/pipelines that are supposed to be able to be run on CBRAIN.

Examples: 
https://portal.conp.ca/pipeline?id=zenodo.3267250
https://portal.conp.ca/pipeline?id=zenodo.3359640

cmadjar:
@liamocn Looks great! One tiny last change. Can the CBRAIN link open on a new tab?

Issue #294, opened 2020-05-28
Number in the dashboard inconsistent between sandboxes and the portal


cmadjar:
There seems to be inconsistencies in the number of datasets graph between sandboxes and the portal but the data should be exactly the same. Need to figure out what is causing those inconsistencies

Screenshot of the graph from the portal:
![Screen Shot 2020-05-28 at 2 48 28 PM](https://user-images.githubusercontent.com/1402456/83181068-59f54f00-a0f2-11ea-8dee-70a78c1a9982.png)

Screenshot of the graph from Cecile's local install:
![Screen Shot 2020-05-26 at 2 20 54 PM](https://user-images.githubusercontent.com/1402456/83181136-72656980-a0f2-11ea-9f5b-5f79442e5fa5.png)

cmadjar:
Suggestions from the CONP dev meeting:
- for datasets, look at the timestamp of the initial commit as the date the dataset was added
- for pipelines, use the information present in the latest Boutiques schema to show the date of pipelines

glatard:
(just to be more accurate, for pipelines this info isn't present in the boutiques schema but it can be retrieved using `bosh search -v`).

PR #298, opened 2020-06-01
Fix issue where the dashboard chart added earlier months unexpectedly

liamocn:
## Related issues

#294 

## Checklist

- [ ] **DO** Unit tests pass.

## Purpose

This addresses one aspect of issue #294 where the x-axis would include months where we have no dataset data. Months are only added if skipped after the first month we have datasets
 
- [ ] Yes
- [ ] No


cmadjar:
The issue comes from the fact that we are relying on `date added` compiled when creating the database of datasets. This will be an issue if we need to recompile the database at some point as the history will be lost.

Instead we should probably use a few fields in the DATS model so the Release Date can be registered? To be discussed at the CONP dev meeting tomorrow...

PR #317, opened 2020-06-16
Add child datasets

liamocn:
## Related issues

#230 

## Checklist

- [x] **DO** Unit tests pass.

## Purpose

This adds tags on the dataset page for and child datasets for that dataset.
 
- [ ] Yes
- [x] No

## Implementation Detail

The portal database now includes a table for parent/child relationships between the datasets. NOTE: `flask db upgrade` and `flask update_datasets` must be ran in order to create and populate this table. Without it, child datasets will not be seen.


jbpoline:
@liamocn - just checking: subdatasets (ie git submodules) are actually semantically "parents" (the derived - or child - dataset is at the top and points to the datasets on which it depends, ie the parents). Are you using child in this way ? it would be good to keep the vocabulary consistent

liamocn:
@jbpoline Yes, I believe this is the case. In fact, the only indicator of the relationship is whether a dataset has a `derivedFrom` attribute in its dats.json. Child datasets are determined by reading all of the `derivedFrom` ahead of time and creating records in the portal database that establish the relationship. I'm also open to using a different vocabulary for the relationship. Right now it's 'parent dataset' and 'child dataset'.

liamocn:
@cmadjar @glatard This PR includes handling changes to the dats.json schema. Can one of you review this when you get a chance? The portal search is currently down.

PR #320, opened 2020-06-18
Dataset CircleCI badge links to summary builds page

joeyzhou98:
---
name: Dataset CircleCI badge links to summary builds page
labels: enhancement

---

## Checklist

- [x] **DO** Unit tests pass.
- [ ] **DO** If new feature, created unit test.

## Purpose
Dataset CircleCI badge is now an anchor link to CircleCI build page so that the user can see why the build for the particular dataset is passing or failling
image
 
#### Does this introduce a major change?
- [ ] Yes
- [x] No

PR #321, opened 2020-06-18
Fix issue of search page not displaying when wrong value for conp_status in DATS file

cmadjar:
This was noticed when testing the import of a new dataset. The conp_status field of the DATS.json model of the new dataset was invalid and this resulted in the data search page not loading. 

The code changes suggested in that PR fixes the issue in case we get another invalid DATS file by mistake. 

Note that separately, we will update the dats_validator to check if the DATS file has proper values in that field. See https://github.com/CONP-PCNO/conp-dataset/issues/345

### DIFF
--- a/app/search/routes.py
+++ b/app/search/routes.py
@@ -174,8 +174,14 @@ def dataset_search():
         paginated = elements
 
         if(sort_key == "conpStatus"):
-            order = {'conp': 0, 'canadian': 1, 'external': 2}
-            paginated.sort(key=lambda o: order[o[sort_key].lower()])
+
+            def determine_order(element):
+                order = {'conp': 0, 'canadian': 1, 'external': 2}
+                if sort_key not in order:
+                    return 3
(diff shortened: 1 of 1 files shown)


glatard:
hi @cmadjar, it looks like this will fix the specific issue we faced today, but we should make sure that the portal won't choke on any other kind of invalid DATS file. I feel like the solution to this should be similar to what is currently done in lines 110-115: 
```
try:
   <something that might fail when the dataset is invalid>
except Exception as e:
  continue # assuming this is in a for loop
```
I'm not sure where this should be put exactly, as I don't know the code, but maybe @liamocn could further advise.

cmadjar:
closing this as it seems that @joeyzhou98 did an equivalent fix in his PR #320

cmadjar:
@glatard there is no for loop in the routes.py and the only field in the DATS model that could have caused an issue due to expected value was conpStatus from what I can tell. @liamocn can confirm though

Issue #14, opened 2019-03-11
[Analytics] CONP Analytics - phase 1


paiva:
### Summary of the analytics to be added on the short term from our Matomo server (Feb 23rd, 2021)

- [x] display Matomo widgets directly into the analytics page using iframes #403
- [x] display number of visits to the portal across time #404
- [x] display number of views per datasets/tools #405
- [x] display popular keyword searches #406 
- [ ] display analytics on dataset providers #407

### Old description of the ticket (Mar 11, 2019)

It would be ideal to deploy some analytics tool and some event tracking systems to better understand user behavior and, if possible, find out what it is being done with the data downloaded. 

Event tracking tags should be consistent with those to be used by the LORIS platform so that we are consistent across both platforms.

Some event tracking tools include:
* Google Analytics
* Snowplow Analytics
* Piwik
* Google WebMaster Tools

Some dashboard tools include:
* Google Analytics
* Grafana

paiva:
@shots47s What is CBRAIN doing for tracking analytics?

shots47s:
@paiva, we use our own internal mechanisms that are baked into the server code.  I would not recommend doing that, I would use something that is canned and out of the box.  I have used GA before, and it works really nice, I know some may have some issues with it being Google, but you can close off sections of the website to the analytics.  I don't have experience with the others.

maltheism:
Hey @paiva, as you know there is discussion of adding GA to LORIS. So let's stay in sync. We can definitely limit what pages GA has access over by not loading it on certain pages.

xlecours:
Matomo (formerly known as Piwik) is free as in freedom (GPLv3)

cmadjar:
Discussed at CONP dev meeting on May 20th, 2020.

What would be next highest priority analytics?
- analytics to display on the portal
  - how many datasets available
  - number of tools available
  - more popular datasets on the portal

- analytics to collect for reviewers and Brain Canada
  - how many datasets visited, how many datasets downloaded etc...?
  - number of users, visits etc..

for next week, prepare a document with metrics of interests.

glatard:
Idea for a metric:
[ ] Number of git clones of the `conp-dataset` repo: https://github.com/CONP-PCNO/conp-dataset/graphs/traffic

cmadjar:
A few ideas of metrics (mostly basic I guess but always good to write them down):
1. [ ] **Number of datasets and tools**
[...]

PR #325, opened 2020-07-15
Add modality/tags charts, analytics page

liamocn:
## Related issues

#14 

## Checklist

- [ ] **DO** Unit tests pass.

## Purpose

This adds two more chart types to the dashboard chart and a dropdown to toggle between them. An analytics page containing all charts has been added
 
- [x] Yes
- [ ] No

## Implementation Detail

The way charts work on the portal had been reworked using the official React bindings for Highcharts.


liamocn:
Yes those bubble graphs move around a bit when they appear. That's by design but I can look into making them appear more static. You can view the names of the keywords by hovering your mouse over the bubbles. I'll look at how I can make that more obvious. Maybe if those charts were bigger overall it would be easier to see.

liamocn:
@cmadjar Hmm, this library that we use for the charts doesn't seem to be capable of doing any of that. I'll have to research new libraries and see if there's one that can be implemented like this.

cmadjar:
@liamocn feedbacks from today's demo:
- Functional links on the bubble if possible to have that
- Remove outer light circle and extend the graph to use more of the white space
- Modify title to something like: `Dataset Modalities Representation By Keywords`/ `Pipeline Representation By Tags`

Hopefully I did not miss any...

cmadjar (changes requested):
@liamocn  thank you for updating this! A few comments:

- there is a lot of white space around the bubble charts (for both the dashboard and the analytics page). Is there a way to decrease the white space?
Screen Shot 2020-10-20 at 2 11 04 PM
- Change titles of graphs to:
  - Dataset Modalities Representation By Keywords
  - Pipeline Representation By Tags
- Given that the title of the graph already mention that the graph represents dataset modalities, maybe the bubble legend is not necessary?
- for the chart, I see histology (2) and Histology (3) but when you click on the bubbles, you get 5 results in the search page => I think histology and Histology should be only one bubble in the chart to match the behaviour of the search page (bubble should not be case sensitive to the tags)

Thank you! Let me know when ready for another round of testing/review.

cmadjar (approved):
Thank you @liamocn !

Issue #283, opened 2020-05-14
Download button


cmadjar:
## Purpose

Have a download button for easier download.

## Proposed download workflow (from the WIKI page that will be removed in favour of this issue)

This draft specifies a data download mechanism from the CONP portal. This mechanism would be an alternate to using the DataLad client in situations where it cannot be easily used.

The following download workflow is proposed:
1. User clicks on dataset download link
2. Browser queries the server for list of URLs in the dataset
3. Server uses DataLad/git-annex to retrieve the list of URLs associated with the dataset files. As this operation might take some time, the use of a caching mechanism is suggested
4. Client downloads the URLs returned by the server

This workflow has the following desirable properties compared to a centralized solution where the portal would download all the data and serve it to the users:
1. No network performance bottleneck in the portal
2. No need for extensive data storage in the portal
3. No need to store user credentials in the portal
4. Possibility to track downloads in the portal

Technical notes:
1. Downloading multiple files from the browser requires the use of web-workers.
2. Cross domain data access requires that the data server authorizes cross-domain queries. This will have to be configured in data providers. 
3. We should make sure that the URLs associated with the data files use a protocol that can be accessible by a browser. It is proposed to agree on a list of protocols (e.g., http, ftp, grid FTP, s3) that the dataset can use, and to specify the list of protocols used in the DATS model.
4. The user should be warned about the size of the dataset when it's too big. Size can be obtained from the DATS model.
5. The browser should group files in an archive when there are many
6. User should be able to select/download only a subset of the files
7. If storage space is available, the portal could cache some files on the server and serve them directly
8. Downloader could send a download confirmation and summary to the portal for tracking purposes

Feedback from DataLad developers. Suggestion: look how OpenNeuro does it:
  - https://github.com/OpenNeuroOrg/openneuro/blob/eff90d7e7fd61b858d43d8f2c73918eff0a32ac4/packages/openneuro-app/src/scripts/datalad/download/download-dataset.jsx
  - https://github.com/OpenNeuroOrg/openneuro/blob/db32c348b991aa3b4e98c88bb51d5c9ffc084019/packages/openneuro-app/src/scripts/datalad/download/download-sw.js#L59:30

Received feedback and approval from TSC (discussed at Feb 3rd, 2020 meeting)

glatard:
See initial specification at https://github.com/CONP-PCNO/conp-portal/wiki/Data-download-mechanism-from-CONP-portal

cmadjar:
[...]

PR #335, opened 2020-09-18
Download button for datasets

xlecours:
## Related issues
https://github.com/CONP-PCNO/conp-portal/issues/283

### DIFF
--- a/app/search/models.py
+++ b/app/search/models.py
@@ -1,5 +1,5 @@
 import datetime as dt
-from functools import lru_cache
+from functools import lru_cache, reduce
 import os
 import json
 import re
@@ -9,6 +9,8 @@ import dateutil
 import requests
 
 from app.models import Dataset
+from datalad import api
+from datalad.api import Dataset as DataladDataset
 
 
 @lru_cache(maxsize=1)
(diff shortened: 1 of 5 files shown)


xlecours:
This is drawing the line between frontend and backend.
@liamocn As long as a GET request is sent to `/download_content` with `id` populated and `version` optionaly populated, I will make the caching system work.... hopefully
Would you want to make the frontend pretty?

xlecours:
@liamocn yes. not authenticated should not be a problem. not too large still needs to be defined.

xlecours:
My last commit (1eaf45e) allows to download any dataset.

app/pipelines/routes.py
```diff
@@ -94,7 +94,7 @@ def pipeline_search():
         ]
 
     # filter out the deprecated pipelines
-    elements = list(filter(lambda e: (not e["DEPRECATED"]), elements))
+    elements = list(filter(lambda e: (not e.get("DEPRECATED", None)), elements))
```
xlecours, line 97: This is unrelated. But fixes an error that prevent the pipeline page to load.

app/search/models.py
```diff
@@ -1,6 +1,7 @@
 import datetime as dt
-from functools import lru_cache
+from functools import lru_cache, reduce
```
xlecours, line 2: reduce is not used.

app/search/models.py
```diff
@@ -42,6 +45,135 @@ def get_latest_test_results():
...
+        return self.maxSize - self.usedSpace
+        
+    @property
+    def cachedDatasets(self):
+        return dict(
+            (f.name, f)
+            for f in os.scandir(self.current_app.config['DATASET_CACHE_PATH'])
+        )
+
+    def getZippedContent(self, dataset):
```
xlecours, line 76: should take version. But I don't know how to get a previous version of a dataset. 
By commit# or tag. Are we using those?

PR #354, opened 2020-10-16
Add matomo tag

liamocn:
## Related issues

#14 

## Checklist

- [x] **DO** Unit tests pass.

## Purpose

This adds the matomo tag to all of the portal pages, for tracking page views, etc

- [ ] Yes
- [x] No

## Implementation Detail

The matomo server URL and site ID are configurable. They will default to `matomo-dev.acelab.ca` and `1` respectively, but can be set in the `.flaskenv` file. This is to distinguish production tracking from any other environment. See the flaskenv template for details.


flaskenv.template
```diff
@@ -17,4 +17,6 @@ ORCID_OAUTH_CLIENT_SECRET=<ORCID SECRET ID>
 ORCID_TOKEN_URL=https://orcid.org/oauth/token
 ADMIN_EMAIL = <Your gmail>
 WEBHOOKS_SECRET=<github webhook secret>
+MATOMO_SERVER_URL=matomo-dev.acelab.ca
```
cmadjar, line 20: can you modify this to 
```suggestion
MATOMO_SERVER_URL=<your Matomo URL>
```

config.py
```diff
@@ -65,6 +65,10 @@ class Config(object):
     # GitHub webhooks
     WEBHOOKS_SECRET = os.environ.get("WEBHOOKS_SECRET")
 
+    # Matomo
+    MATOMO_SERVER_URL = os.environ.get("MATOMO_SERVER_URL", "matomo-dev.acelab.ca")
```
cmadjar, line 69: can you modify this to

```suggestion
    MATOMO_SERVER_URL = os.environ.get("MATOMO_SERVER_URL", <your Matomo URL>)
```

cmadjar (changes requested):
@liamocn two minor comments. Otherwise looks good.

config.py
liamocn: The fallback value needs to be valid so I've removed it from here instead

cmadjar (approved):
LGTM

PR #362, opened 2020-10-29
Read URL of CBRAIN from JSON file instead of hardcoding it to login URL

cmadjar:
Instead of hardcoding the CBRAIN URL to the login page when clicking on the CBRAIN buttons, this will read the URL from the Boutiques descriptor of the tool instead in case the URL contains the pipeline ID so that users can be redirected directly to the tool page on CBRAIN.

### DIFF
--- a/app/pipelines/routes.py
+++ b/app/pipelines/routes.py
@@ -152,7 +152,7 @@ def pipeline_search():
                 if "cbrain" in url:
                     element["platforms"][0]["img"] = url_for(
                         'static', filename="img/run_on_cbrain_green.png")
-                    element["platforms"][0]["uri"] = "https://portal.cbrain.mcgill.ca"
+                    element["platforms"][0]["uri"] = url
                 else:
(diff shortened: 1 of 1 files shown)


app/pipelines/routes.py
```diff
@@ -152,7 +152,7 @@ def pipeline_search():
                 if "cbrain" in url:
```
xlecours, line 152:
```suggestion
                if url.startswith('https://portal.cbrain.mcgill.ca'):
```

app/pipelines/routes.py
```diff
@@ -260,7 +260,7 @@ def pipeline_info():
             if "cbrain" in url:
```
xlecours, line 260:
```suggestion
            if url.startswith('https://portal.cbrain.mcgill.ca'):
```

xlecours (changes requested):
for security, let's make sure it starts with cbrain's portal url.

xlecours:
I tested with the suggested changes and it works.
The descriptors (other than Zenodo.4148091) could be changed now.

PR #364, opened 2020-10-30
Data download caching

xlecours:
https://github.com/CONP-PCNO/conp-portal/issues/283#issuecomment-693443035

This is working but when it fails, it is not graceful for the least.
We should handle errors when the backend can't download the dataset (missing files, subdatset, zenodo zip file, ... )

Known issues:
 - datalad export_archive fails when a subdataset is empty. (ex: https://portal.conp.ca/dataset?id=projects/BigBrain_3DClassifiedVolumes)
 - This solution works for sequential request. What happen when many users download file at the same time. And what happen when they try to download the same file ??? There should probably be a lock mechanism to isolate datalad.install and datalad.drop calls. Should our max_cache_size be multiplied or devided by the number of concurrent requests?


app/templates/dataset.html
```diff
@@ -97,21 +103,23 @@ <h2><span style="color:red;">CONP Portal</span> | {{ data.title }} </h2>
   {% endif %}
 </div>
 
+{% if not showDownloadButton %}
```
xlecours, line 106: @liamocn I think we can still show "alternate" download instruction even when the download button is there.
@cmadjar ?

liamocn: I'm open to that, if that's the decision. Maybe just change the section heading if there's a download button instead?

cmadjar: I agree with Xavier, we should keep the instructions as alternate download instructions.

Question: if there is an error when clicking on the download button, does the user get a message? For example, if there is a `datalad get` error?

liamocn: Hi @cmadjar @xlecours . I've made these changes in a PR mentioned below. The download instructions are back and there is both a progress view and an error view to provide feedback to the user.

cmadjar:
@xlecours could you please review and merge the PR @liamocn sent to your branch? https://github.com/xlecours/conp-portal/pull/2

There are also conflicts that need to be resolved. Thanks!

cmadjar:
See https://docs.google.com/document/d/1secjwOXBjk0G3aOi6-_k2Euszir-vyEwxIIiO2Gb6jg/edit for proposed plan on how to implement this

xlecours:
I messed with git. This PR is now at https://github.com/CONP-PCNO/conp-portal/pull/381

Issue #155, opened 2020-01-17
Only look for datasets in projects and investigators directories 


xlecours:
`app/cli.py::_update_datasets` prints this error:

> [ERROR  ] DATS.json file can`t be found in conp-portal/data/conp-dataset/scripts/dats_validator/conp-dats.

xlecours:
investigator directory do not exists anymore

PR #365, opened 2020-11-11
Update `update_datasets` script

liamocn:
## Related issues

#155 

## Checklist

- [ ] **DO** Unit tests pass.

## Purpose

This PR ensure the `update_datasets` script skips directories that are not in either the `/projects` or `/investigator` directories. An ERROR message was being thrown every time this script ran because of searching for DATS files where they don't exist.

## Current behaviour

All subdirectories are searched for a DATS.json file including those where none are expected

## New behaviour

Only the `/projects` and `/investigators` directories are searched 
 
- [ ] Yes
- [ ] No

### DIFF
--- a/app/cli.py
+++ b/app/cli.py
@@ -210,8 +210,8 @@ def _update_datasets(app):
                 print("\033[0m")
                 continue
 
-        # The folowing relates to the DATS.json files of the projects. Skip directories that aren't projects or investigators.
-        patterns = [app.config['DATA_PATH'] + '/conp-dataset/projects/*', app.config['DATA_PATH'] + '/conp-dataset/investigators/*']
+        # The folowing relates to the DATS.json files of the projects. Skip directories that aren't projects.
+        patterns = [app.config['DATA_PATH'] + '/conp-dataset/projects/*']
         if not any(fnmatch.fnmatch(ds['path'], pattern) for pattern in patterns):
             continue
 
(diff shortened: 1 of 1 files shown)


app/cli.py
```diff
@@ -209,6 +210,11 @@ def _update_datasets(app):
                 print("\033[0m")
                 continue
 
+        # The folowing relates to the DATS.json files of the projects. Skip directories that aren't projects or investigators.
+        patterns = [app.config['DATA_PATH'] + '/conp-dataset/projects/*', app.config['DATA_PATH'] + '/conp-dataset/investigators/*']
```
xlecours, line 214: you can remove investigator. It doesn't exists anymore. 
If we ever add it back we will change this line.

xlecours (changes requested):

cmadjar (changes requested):
@liamocn this is failing Travis. If you could also integrate Xavier's comment at the same time, that would be awesome. Thanks!

app/cli.py
```diff
@@ -209,6 +210,12 @@ def _update_datasets(app):
                 print("\033[0m")
                 continue
 
+        # The folowing relates to the DATS.json files of the projects.
```
cmadjar, line 213:
```suggestion
        # The following relates to the DATS.json files of the projects directory in the conp-dataset repo.
```

cmadjar (changes requested):
one tiny change on the comment. Otherwise, looks good!!

cmadjar (approved):
LGTM

PR #381, opened 2020-12-02
Download button caching

xlecours:
This adds a dataset cache cofiguration which is scanned for existing zipped files for each datsets. If the file exists, the dataset will be downloadable via the frontend.

It also change the nginx config to serve large files directly. 

Once merged, will need to:
- add the lines in /etc/nginx/sites-available/conp-portal
- `sudo service nginx restart`


app/search/models.py
```diff
@@ -43,6 +43,43 @@ def get_latest_test_results():
...
+           'gz'
+        ])
+
+        cached = self.cachedDatasets.get(zipFilename)
+        zipped = cached.path if cached is not None else None
+
+        if zipped is None:
+            raise RuntimeError('Not found')
+
+        os.system('touch -a ' + zipped)
```
cmadjar, line 80: not sure I understand this line. Was it for testing purposes and it should be removed?

cmadjar (commented):
@xlecours one tiny comment

app/search/models.py
xlecours: This is updating the 'last accessed timestamp' of the file on the filesystem.

xlecours: It was used when I was deleting the least recently accessed. I forgot to remove the line.

app/search/models.py
```diff
@@ -43,6 +43,43 @@ def get_latest_test_results():
...
+    def getZippedContent(self, dataset):
+        """
+          1. Server checks if a zip file already exists for this version.
+          2. Return zip file
+        """
+
+        datasetmeta = DATSDataset(dataset.fspath)
+        zipFilename = '.'.join([
+            datasetmeta.name.replace('/','__'),
+            datasetmeta.version,
```
cmadjar, line 69:
```suggestion
            '_version_',
            datasetmeta.version,
```

cmadjar: suggestion to add the string `_version_` between the dataset name and the dataset version in the name of the tar.gz file

cmadjar (commented):
@xlecours I cannot see a download button on the portal for the example dataset I wanted to test.

Here is what I have:
- `.flaskenv` updated to add `DATASET_CACHE_PATH=${HOME}/.cache/conp-dataset`
- In `/Users/cmadjar/.cache/conp-dataset`, I created `Expression_based_polygenic_score_from_the_amygdala_5HTT_gene_network1.0.tar.gz`

Is it working for you?

app/search/models.py
```diff
@@ -43,6 +43,43 @@ def get_latest_test_results():
...
+           'tar',
+           'gz'
+        ])
```
cmadjar, line 72:
```suggestion
        zipFilename = datasetmeta.name.replace('/','__') + '_version-' + \
                        datasetmeta.version + '.tar.gz'
```

app/static/css/core.css
```diff
@@ -162,7 +162,6 @@ Colour names courtesy of chir.ag/projects/name-that-color
 
 .btn {
   padding: 1rem;
-  width: 100%;
```
cmadjar, line 165: This should not be removed as this affect also the terms of use buttons and the contextual help buttons.

cmadjar (changes requested):
@xlecours I did some more testing and noticed that when the download button is being displayed, all the other metadata is not shown anymore, just the version. This needs to be fixed.

In summary, need to:
- [ ] fix the bug that the metadata of the dataset are not shown anymore when the download button appears (except for the version)
- [ ] put back the width: 100% for the .btn since this also affects the terms of use window and will affect the contextual help windows that are now on the portal
- [ ] change the way the archive file is named as suggested in my comments today (forget the old suggestion I made yesterday, it would not have worked)

Once those fixes are done, then the download button should work on the portal. Most datasets that could be downloaded have been prezipped.

Thank you!!

xlecours:
@cmadjar

xlecours:
@cmadjar Changes pushed. I am not working today. If you want to take a look at it, go ahead. Otherwise I'll test it on `dev` Monday.

xlecours:
I think discussion needs to happen about the Download button location the page and if we want the autoindex opton for `/data/`

cmadjar:
@xlecours did you pulled the latest code of conp/master? The dataset page have been redesigned and there is a download button in the card of the dataset in the search page and dataset page. Right now, it links to the download documentation but for the datasets that could be downloaded, it could use your latest download code instead.

Thoughts?

PS: I am working on LORIS tasks today and am not sure I will have time to look at your changes today or tomorrow but will definitely do on Wednesday PM.

xlecours:
> @xlecours did you pulled the latest code of conp/master? The dataset page have been redesigned and there is a download button in the card of the dataset in the search page and dataset page. Right now, it links to the download documentation but for the datasets that could be downloaded, it could use your latest download code instead.
> 
> Thoughts?
> 
> PS: I am working on LORIS tasks today and am not sure I will have time to look at your changes today or tomorrow but will definitely do on Wednesday PM.

Yes, I have the latest code from the master branch. I wasn't sure about the desired behaviour. Since the download button at the top links to the download instruction, I thought the "real download button" should be in the download section. Inm our previous dev, there was a section that conditionally appear for that download button. Should it be brought back?

app/templates/dataset.html
```diff
@@ -218,8 +218,25 @@ <h5><strong>Description: </strong></h5>
...
+  <div class="d-flex flex-row align-items-center">
+    <div class="col-md-2">
+      <a href="{{ zipLocation }}" class="btn btn-outline-primary" id="downloadButton">
+        Download this version
+      </a> 
+    </div>
+  </div>
+  <div class="d-flex flex-row align-items-center">
+      {% endif %}
+      <h2 class="px-2">Dataset Download Instructions</h2>
```
cmadjar, line 239:
```suggestion
      <h2 class="px-2">Dataset Download Instructions With DataLad</h2>
```

cmadjar:
@xlecours I sent the following PR to your branch: https://github.com/xlecours/conp-portal/pull/3 to:
1) update the code to latest version of CONP-PCNO/master
2) have the Direct Download section always display but if a dataset cannot be downloaded through the frontend, then the button is deactivated and there is a text explaining the user why the dataset cannot be downloaded.

I pulled that code onto https://portal-dev.conp.ca so you can have a look at what it looks like when you get back.

cmadjar (approved):
Looking great! Thank you @xlecours :)

PR #386, opened 2020-12-11
Fixed the dashboard links to the three PREVENT-AD datasets

cmadjar:
This fixes the links in the dashboard to the preventad-open-bids and preventad-registered datasets that were pointing to the preventad-open instead.


liamocn:
LGTM!

Issue #372, opened 2020-11-19
Page navigation buttons does not redirect to the top of the page list after clicking on it


cmadjar:
## Purpose

At the bottom of the screen of the data and pipeline search pages, when clicking on the next page to be displayed, the user stays at the bottom of the page instead of being redirected to the top of the page.

Issue #371, opened 2020-11-19
Add an 'all' category in the drop down "number of datasets/pipelines displayed per page"


cmadjar:
## Purpose

In the suggestion of number of datasets/pipelines to display per page in the search and pipelines page, it would be nice to add an 'all' option that would list all the datasets/pipelines into a single page.

Current available options are 5, 10, 15, 20

Issue #353, opened 2020-10-14
Losing data search filter when going back from a dataset page


cmadjar:
User story:
- user filters for datasets using the search box and applies filter (Neuroimaging)
- user clicks on a dataset to open the dataset page
- user clicks on the back arrow of the browser expecting to see the same filter (Neuroimaging) but instead user has to reenter their filters in the search box

Is there a way to keep the search filter information so that when the user clicks on the back button, their filter remains valid?

Would also be good to retain the information about the "number of datasets/pipelines to display per page" when going back to the search page.

cmadjar:
Note from the meeting of November 19th, 2020: this might be a bit tricky to implement given the flask architecture. @liamocn will investigate how this could be done.

PR #392, opened 2020-12-29
Search page fixes to backstack, pagination and added 'View all results'

liamocn:
## Related issues

#353
#371 
#372 

## Checklist

- [ ] **DO** Unit tests pass.

## Purpose

NOTE: Fixes for issues 371 and 372 were added to this PR as they both concern the search page, just like 353.

Issue #353 
NOTE: This was only a problem on Chrome since they haven't implemented a forward/backward cache like others have: https://stackoverflow.com/questions/60382446/which-browsers-have-a-back-forward-cache-a-bfcache-keeps-modifications-to-dom
This meant that on navigating back in Chrome, the browser would perform a refresh and the search page would lose state.
This PR uses the browser query parameters to store the search state so that it will survive any page refresh (via back navigation or otherwise)

Issue #371 
A 'View All' option has been added to the 'results per page' dropdown on the search page. This will simply display all of the results in a single page

Issue #372 
When the datasets on the search page load, we now scroll the window to the top so that the user sees the first result. This is mostly useful for changing page, but will take effect whenever the search page loads new results because of a user action.

- [ ] Yes
- [ ] No


app/pipelines/routes.py
```diff
@@ -140,17 +146,16 @@ def pipeline_search():
     # else take the first one and set logo
     # TODO right now, this handles CBRAIN and one other platform
 
-    #Mandana issue 378
+    # Mandana issue 378
```
cmadjar, line 149:
```suggestion
    # read the list of pipelines also available in CBRAIN
```

cmadjar: the comment above was not clear. I know this is not technically part of your PR but it might be good to update it to a more useful comment anyway since you are changing the line.

cmadjar (changes requested):
Thank @liamocn! I just tested everything and found the following that would need to be resolved.

- [x] View all results -> works great! The only thing is that I would change the drop down label to 'All' instead of 'View All'
- [x] backstack -> works great for modality and file format in the data search page
  - [x] sort by option is not kept
  - [ ] the search box behaves strangely (both data and pipelines pages). The page jumps every time we enter a letter in the search box and we lose focus
  - [ ] the backstack of the pipelines page redirects to the data search page instead of the pipelines search page
- [ ] pagination -> there is a significant delay before the user gets redirected to the top of the search result

cmadjar:
@liamocn thank you for the fixes! The only thing I notice is that we lose the sorting option when we go back to the search page from a tool or a dataset page. Would it be hard to implement or should it be a separate PR?

liamocn:
> 
> 
> @liamocn thank you for the fixes! The only thing I notice is that we lose the sorting option when we go back to the search page from a tool or a dataset page. Would it be hard to implement or should it be a separate PR?

@cmadjar What browser are you using? I can't reproduce that at all...

cmadjar:
@liamocn Chrome

liamocn:
My bad. Should be fixed now.

cmadjar (approved):
Working great! Thank you @liamocn

PR #394, opened 2021-01-15
File format drop down harmonization and sorting

cmadjar:
This harmonizes the format drop down filter with the format convention in place in the DATS.json files describing the dataset. In addition, the list of file format is sorted in the drop down.

Before:
Screen Shot 2021-01-15 at 12 41 24 PM

After:
Screen Shot 2021-01-15 at 12 40 39 PM

## Related issues

https://github.com/CONP-PCNO/conp-dataset/issues/491

### DIFF
--- a/app/search/routes.py
+++ b/app/search/routes.py
@@ -203,9 +203,11 @@ def dataset_search():
         if request.args.get('formats'):
             filterFormats = request.args.get('formats').split(",")
             elements = list(
-                filter(lambda e: e['format'] is not None, elements))
-            elements = list(filter(lambda e: all(item in (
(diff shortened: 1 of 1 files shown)


liamocn (changes requested):
Hey @cmadjar, this looks great but the sorting by format on the search page doesn't work anymore. This is because the formats don't match what's in the DATS.json files. What the user sees in the formats list will have to match what's in the DATS files.

cmadjar:
@liamocn thank you for pointing that out. This should be fixed now.

liamocn (approved):
Looks good!

PR #399, opened 2021-01-22
Add GIfTI to the list of exceptions for file formats to handle in the drop down of the dataset search page

cmadjar:
While recrawling the BigBrain dataset, we noticed that GIfTI should be added to the list of file format exceptions that do not follow the upper case convention.


cmadjar:
@liamocn when you get a chance if you could review this, that would be great. Thank you!

liamocn (approved):
Looks good!

Issue #375, opened 2020-11-20
Dataset page improvements


cmadjar:
This is related to issue #186. When tackling that issue, we realized that it would be a good opportunity to redesign/revamp the dataset page. So far, the following improvements have been discussed:

- refactoring of the card summarizing the dataset content (see #374)
- links are currently displayed in blue which attracts the user's eyes on information that are not necessarily the information we want to focus on (like link to external source databases). Need to modify the colour of the links to either the same colour as the rest of the text or lighter colour.
- figure out a way to better organize the different sections of the page (important info in the card, more DATS info on the dataset, download instructions, dataset README)

cmadjar:
@3design I am assigning this to you for now to finish up the mock up first, then I will assign it to @liamocn once we are happy with the mock ups. If you could update this ticket with things I missed or progress, that would be great. This way we keep everything in one place. Thank you :)

cmadjar:
Link to the final mock up of @3design : https://drive.google.com/file/d/1GSndNuIF9EUteqSMENYSKiQiEgo3DUAS/view

Issue #374, opened 2020-11-20
Refactoring of the dataset card in the search list and dataset page


cmadjar:
## Purpose

The dataset page needs some refactoring and re-design. Preliminary work on this can be found at the following link that Derek gave us: https://drive.google.com/drive/folders/1db1jsWa3dn787mUtN4-mlUkalmKuXJP1?usp=sharing

The general idea behind this new card layout would be to have 3 divs in the card: 
- one for the logo and country/views/downloads/likes icons; 
- one for the dataset information (title DATS info etc...)
- one for the action buttons (download metadata, download dataset...)
```
<div class="card">
  <div class="logo_and_metrics">
    <img>
    <icon>
    <metric>
  </div>
  <div class="dataset_info">
    <h1>title</h1>
    <who>
    <when>
    <what>
    <what details>
  </div>
  <div class="downloads">
    <h2>Download</h2>
    <button>Metadata</button>
    <button>Dataset</button>
    <icon>conditional access</icon>
   </div>
</div>
```

The dataset info section section would be organized this way:
- Title: This is self explanatory and can word-wrap as needed
- Who: This would be data pulled from the dataset that describes who was responsible for the creation, including institution etc. I would suggest for the search page card, we pull the first name from the dataset, and add 'et al.' as needed. On the dataset page itself, we can display all people listed
- When: This is straightforward - when data was added and last updated
- What: This shows top level information about the dataset such as the version, data type(s), modality, license, etc
- What details: This is finer details about the dataset itself such as file size, file formats, # of files, etc.

In addition, the dataset description from the DATS file could be displayed in a fourth div with a certain limit of characters and a link to expand the text on the card if the user is interested in reading the whole description.

Outside of the cards, i would suggest a few small tweaks to the search page:
- less white space on the sidebar navigation 
- make form fields and drop down buttons the SAME HEIGHT :)
- label the modalities and file types with 'filter results by' to improve ux

cmadjar:
@3design I am assigning this to you for now to finish up the mock up first, then I will assign it to @liamocn once we are happy with the mock ups. If you could update this ticket with things I missed or progress, that would be great. This way we keep everything in one place. Thank you :)

cmadjar:
Link to the final mock up of @3design : https://drive.google.com/file/d/1GSndNuIF9EUteqSMENYSKiQiEgo3DUAS/view

@liamocn do you have everything you need to start working on the dataset pages?

Thank you both!

Issue #255, opened 2020-04-10
Dataset browse link


glatard:
## Purpose

We could have a "browse" button in the dataset page that would link to the dataset repo on GitHub. For instance, for the BigBrain dataset the link would currently point to [here](https://github.com/conpdatasets/bigbrain-datalad/tree/792702f93455fcd565e76026a81e2e71fb46d444).

## Context

It would allow users to browse a dataset before downloading it.

## Possible Implementation

Add a browse button, to the dataset page (the page that shows once we cliked on the dataset in the dataset list), that button should open the dataset repo in a new tab.

PR #411, opened 2021-02-16
Dataset page updates

liamocn:
## Related issues

#374 #375 #255

## Checklist

- [ ] **DO** Unit tests pass.

## Purpose

This implements the new designs for the dataset search and dataset pages.
 
- [ ] Yes
- [ ] No

## Implementation Detail

NOTE: The new design requires a link to the remotes however they are not being stored anywhere currently. You must run

`flask db upgrade`

followed by

`flask update_datasets`

to pull the remote URLs of the datasets into the database.


cmadjar (changes requested):
ooooooh, this is looking so good!! Love the chips on the page that automatically lead to filtered search content. It looks so great! Thank you for doing this!!

@liamocn a few things I noticed while testing:
- I get this warning when I run `flask update_datasets`
```
[WARNING] realpath of PWD=/Users/cmadjar/Github/CONP-PCNO/conp-portal is /Users/cmadjar/Github/CONP-PCNO/conp-portal whenever os.getcwd()=/Users/cmadjar/GitHub/CONP-PCNO/conp-portal. From now on will be returning os.getcwd(). Directory symlinks in the paths will be resolved
```
- the modality and file format filters should be sorted alphabetically
- the file format filter, I see 2 H5, NIFTI and NIfTI while there should be only NIfTI, GIFTI should actually be written GIfTI, CSV present twice, DICOM present twice. There might be other duplicates that I missed.
- when displaying all rows, the navigation arrows at the bottom of the table should be hidden
- missing comma and space between modalities listed in the cards when there are more than one
- weird layout of the PREVENT-AD dataset card. Is there a way to improve that? For example, when there are more than X values for modality, only display the first X ones followed by `...` and the user could see the content when mouseover?
- for greyed out buttons, I wonder if we should add a text when the user mouse over: "this dataset is not available for download because of its large size. We offer download instructions via datalad on the dataset page." The text probably needs to be improved but you get the idea.
- the description part of the dataset page looks kind of funny for most dataset due to small content in the description. I wonder if it would not be better to move the Description header to the left and have the header display like the title of a dataset instead? Just wondering. I have a hard time visualizing if that would be better.
- regarding the fields displayed in the dataset page, I think it was mentioned that more of the DATS content could be displayed under a More section that would expand if users wanted to see more fields. Fields to be displayed could be reviewed with @emmetaobrien.

cmadjar:
PS: @liamocn the PR fails travis

liamocn:
@cmadjar Those points should be resolved now. Regarding the last one, I can show tabulated data (as is now the case) but I'm not sure exactly what to put in the table. We could show everything from just a few selected items to the whole DATS if we wanted to.

cmadjar (changes requested):
@liamocn thank you for all the fixes!

- I still get this warning when I run `flask update_datasets`. Do you have it as well?
```
[WARNING] realpath of PWD=/Users/cmadjar/Github/CONP-PCNO/conp-portal is /Users/cmadjar/Github/CONP-PCNO/conp-portal whenever os.getcwd()=/Users/cmadjar/GitHub/CONP-PCNO/conp-portal. From now on will be returning os.getcwd(). Directory symlinks in the paths will be resolved
```
- File format filter: sorting appears to be case dependent. (bigWig at the end of the list)
- Thank you for fixing the card layout! Much better :). I am just wondering whether Version should be moved on the same line as date added, date updated and Modalities be its own line. The date/version line gives info about the state of the dataset and modalities gives a bit of description about what is in the dataset. To keep the idea of Derek to have related information in same line when possible. I don't know what that would look like though, so if you think the date/version line might be too clouded, no worries.
- for the disabled download button, the text is "disabled tool tip". Question for you, is it just an anchor for where the download button will be? If so, since we do not have yet a working version of the download button, I wonder if it would not be best to modify the label to `Dataset Download Instruction`, or if too long, keep it as Dataset and have the button redirect to the dataset download instructions for now. When the actual downloading through frontend works, we can relink it to the download algorithm.
- number of subjects should also be displayed in the dataset page source. Only keep one of those to remain consistent
- in card landing page is labelled origin but in the dataset details it is labelled  
- for the extra DATS fields to display, here is a list of fields that could be displayed in the more section (no need to repeat what is already there in the detailed of the dataset page)
    - origin fields when they are available
    - dimensions if available
    - produced by if available
    - isAbout if available
    - acknowledges if available
    - spatial coverage if available
    - note: only display the more button if there is actual data to be displayed in the list above
- when reducing the size of the page, the logo gets out of the card: 
Screen Shot 2021-02-25 at 11 40 14 AM

- in the pipeline page, the Boutiques link is getting out of the box
Screen Shot 2021-02-25 at 11 41 37 AM

Thank you @liamocn.

liamocn:
@cmadjar I'm not seeing that console warning when I run `flask update_datasets` but it looks like something from datalad?

Anyway, this is ready for another review when you're ready.

cmadjar (approved):
This all looks great! Thank you @liamocn :-)

There were conflicts with the master branch but they were straightforward to resolve. Maybe just look at the last commit to ensure I did not mess up anything?

cmadjar:
@liamocn this is failing Travis due to more stringent checks merged early. Looks like you just need to remove the `import re` from `app/search/routes.py`

```
$ ./lint.sh
./app/search/routes.py:8:1: F401 're' imported but unused
import re
^
1     F401 're' imported but unused
1
The command "./lint.sh" exited with 1.
```

PR #410, opened 2021-02-15
Query Matomo through its API and insert results into the portal's database tables

cmadjar:
### Description

Analytics are gathered by our instance of Matomo. This PR adds new tables to `models.py` used by flask to create the database tables that will store the Matomo API responses identified in #405, #404 and #406. The information stored in those tables will be used to plot the following analytics graphs:
- number of visits/uniq visitors per week, months over time
- number of views per dataset 
- number of views per pipeline
- popular site search keywords

### How to populate the tables

- 1) create the tables: `flask db upgrade`
- 2) populate the table content by querying Matomo: `flask update_analytics`

##### 1)  `matomo_daily_visits_summary` 
This table will store all the stats reported by Matomo regarding daily visits to the portal. Columns of the table are:
```
    id                   = ID
    date                 = date 
    avg_time_on_site     = average time spent, in seconds, on this page
    bounce_count         = number of visits that bounced (viewed only one page)
    max_actions          = maximum number of actions in a visit
    nb_actions           = number of actions (page views, outlinks and downloads)
    nb_uniq_visitors     = number of unique visitors
    nb_users             = number of unique active users (visitors with a known User ID).
                           If you are not using User ID then this metric will be set to zero.
    nb_visits            = number of visits (30 min of inactivity considered a new visit)
    nb_visits_converted  = number of visits that converted a goal
    sum_visit_length     = total time spent, in seconds
```

##### 2) `matomo_daily_dataset_page_views_summary`
This table will store all the daily stats for each dataset_id. Columns are:
```
    id               = ID
    date             = date
    dataset_id       = dataset_id as stored in datasets table
    url              = page URL
    label            = page label
    nb_hits          = number of views on this page
    nb_visits        = number of visits (30 min of inactivity considered a new visit)
    nb_uniq_visitors = number of unique visitors
    sum_time_spent   = total time spent on this page, in seconds
    avg_time_on_page = average time spent, in seconds, on this page
```

##### 3)  `matomo_daily_get_page_urls_summary`
This table will store all the daily stats for all URLs except for the dataset pages. Columns are: 
```
    id               = ID
    date             = date
    url              = page URL
    label            = page label
    nb_hits          = number of views on this page
    nb_visits        = number of visits (30 min of inactivity considered a new visit)
    nb_uniq_visitors = number of unique visitors
    sum_time_spent   = total time spent on this page, in seconds
    avg_time_on_page = average time spent, in seconds, on this page
```

##### 4) `matomo_daily_site_keyword_searches_summary`
[...]

PR #422, opened 2021-03-10
Make SQL queries mysql compatible

cmadjar:
When running `flask db upgrade` on production after #410 was merged, the upgrade failed due to missing parameter length for String columns. This does not appear to be an issue with SQLITE but turns out to be one on the production MariaDB server.

The changes in this PR will ensure that `flask db upgrade` can be run on production to create the analytics tables.

Note: I tested by setting up a MySQL server on my local instance and I was able to run `flask db upgrade` + `flask update_analytics` with those changes to the schema.


liamocn:
Looks good!

Issue #406, opened 2021-02-03
Display popular keyword searches


cmadjar:
## Purpose

Display on the portal popular keyword searches. 

**_Before implementing that, do some pre-analysis to see if we can get helpful information from that endpoint and how feasible it is to implement._**

## Context

Matomo offers a nice API capability that allows to grep some summary statistics that we could process and display in addition to the widgets that Matomo already offers.

In order to get an idea of keyword searches, we need to perform some analysis on the data collected by Matomo so the easiest would be to import summary statistics into a table in the database of the portal via the API.

## Possible Implementation

Several steps will be required in order to obtain that:

1. Create a table in the flask database that will contain results from the Matomo API query (`matomo_getSiteSearchKeywords` - schema to be determined based on the API response and the information we might need)
2. Create/add to the function `update_analytics` to run with flask that will update the table above
3. Query the Matomo API for a given timeframe (day, week, month available. Start maybe with day) to produce summary statistics on the method `Actions.getSiteSearchKeywords` and insert the results into the `matomo_getSiteSearchKeywords` table
4. Create a function that will query `matomo_getSiteSearchKeywords` and analyse which keyword was used before exiting the search page (to get read of most of the b, bi, big... for bigbrain searches etc...)
5. Display a table or graph showing popular keywords

## Related issues 

#14

cmadjar:
#410 queries Matomo's API and inserts results in `matomo_daily_site_keyword_searches_summary`. Columns are:
```
 id                  = ID
    date                = date
    avg_time_on_page    = average time spent, in seconds, on this page
    bounce_rate         = ratio of visits leaving the website after landing on this page
    exit_nb_visits      = number of visits that finished on this page
    exit_rate           = ratio of visits that do not view any other page after this page
    label               = keyword searched
    nb_hits             = number of views on this page
    nb_pages_per_search = number of pages displayed for the searched keyword
    nb_visits           = number of visits (30 min of inactivity considered a new visit)
    segment             = segment with keyword search
    sum_time_spent      = total time spent on this page, in seconds
```

Query to select the total number of hits per label.
```
SELECT label, SUM(nb_hits) as search FROM matomo_daily_site_keyword_searches_summary GROUP BY label ORDER BY search;
```
My guess is that it is what we wish to display. Results of that query looks like:
```
{ ... more available below 10 hits }
ubc|11
braincode|12
preventad|12
MRI|13
neuromod|13
mri|15
bids|16
mouse|16
[...]

Issue #405, opened 2021-02-03
Display number of views per datasets/tools


cmadjar:
## Purpose

Display on the portal the number of views per dataset/tools. The same mechanism will be used in order to collect and display that information.

## Context

Matomo offers a nice API capability that allows to grep some summary statistics that we could process and display in addition to the widgets that Matomo already offers.

In order to get an idea of the number of views, we need to perform some counts on the data collected by Matomo so the easiest would be to import summary statistics into a table in the database of the portal via the API.

## Possible Implementation

Several steps will be required in order to obtain that:

1. Create a table in the flask database that will contain results from the Matomo API query (`matomo_getPageUrls` - schema to be determined based on the API response and the information we might need)
2. Create a function `update_analytics` to run with flask that will update the table above
3. Query the Matomo API for a given timeframe (day, week, month available. Start maybe with day) to produce summary statistics on the method `Actions.getPageUrls` and insert the results into the `matomo_getPageUrls` table
4. Create a function that will query `matomo_getPageUrls` and count the visited pages per dataset/tools and insert/update that into a new table `dataset_number_of_views`/`pipelines_number_of_views`
5. Display the number of times a dataset has been viewed on the search pages in the card below the dataset/tool logo
6. Optionally, create graphs or table listing ranking of datasets/pipelines in the analytics page

## Related issues 

#14

cmadjar:
#410 queries Matomo's API and inserts results in `matomo_daily_dataset_page_views_summary` for the datasets and `matomo_daily_get_page_urls_summary` for all the other pages of the portal (including pipeline pages). Columns are:

Columns for `matomo_daily_get_page_urls_summary` are:
```
    id               = ID
    date             = date
    url              = page URL
    label            = page label
    nb_hits          = number of views on this page
    nb_visits        = number of visits (30 min of inactivity considered a new visit)
    nb_uniq_visitors = number of unique visitors
    sum_time_spent   = total time spent on this page, in seconds
    avg_time_on_page = average time spent, in seconds, on this page
```

Columns for `matomo_daily_dataset_page_views_summary` are:
```
    id               = ID
    date             = date
    dataset_id       = dataset_id as stored in datasets table
    url              = page URL
    label            = page label
    nb_hits          = number of views on this page
    nb_visits        = number of visits (30 min of inactivity considered a new visit)
    nb_uniq_visitors = number of unique visitors
    sum_time_spent   = total time spent on this page, in seconds
[...]

Issue #404, opened 2021-02-03
Display number of visits to the portal across time


cmadjar:
## Purpose

Show a simple histogram of the number of visits across time on the analytics page.

## Context

Matomo offers a nice API capability that allows to grep some summary statistics that we could process and display in addition to the widgets that Matomo already offers.

## Possible Implementation 

Two possibilities:

1. link the following [widget](https://matomo-dev.acelab.ca/index.php?module=Widgetize&action=iframe&forceView=1&viewDataTable=graphEvolution&disableLink=0&widget=1&moduleToWidgetize=VisitsSummary&actionToWidgetize=getEvolutionGraph&idSite=2&period=range&date=last30&disableLink=1&widget=1
) into the portal but it might have just too much information for the users?
2. collect the numbers via the Matomo API `VisitsSummary.get` endpoint and plot a graph with the number of visits and number of unique visits across time (in months).  

## Related issues 

#14

cmadjar:
Option 2 has been chosen at the CONP dev meeting of February, 10th 2020.

cmadjar:
#410 queries Matomo's API and inserts results in `matomo_daily_visits_summary`. Columns are:

```
    id               = ID
    date             = date
    dataset_id       = dataset_id as stored in datasets table
    url              = page URL
    label            = page label
    nb_hits          = number of views on this page
    nb_visits        = number of visits (30 min of inactivity considered a new visit)
    nb_uniq_visitors = number of unique visitors
    sum_time_spent   = total time spent on this page, in seconds
    avg_time_on_page = average time spent, in seconds, on this page
```

Query to display the number of visits and uniq visitors across time:
```
SELECT date, nb_visits, nb_uniq_visitors FROM matomo_daily_visits_summary;
```

The results of the query would be:
```
{ ... more data available for previous dates }
2021-02-14|3|3
2021-02-15|3|1
2021-02-16|15|12
2021-02-17|28|27
2021-02-18|17|15
2021-02-19|11|9
2021-02-20|7|6
2021-02-21|4|4
```

Then the graphs could display the number of visits per months, weeks or days depending on what the user wishes to see.

cmadjar:
@liamocn the display for this would be a monthly cumulative histogram starting in June 2020 (month of the official release of the portal)

Let me know if there is anything else I can help with in that process. Happy to help creating the routes for that as well. On the React front, I am less useful though ;)

Issue #403, opened 2021-02-03
Display Matomo widgets into the analytics page


cmadjar:
## Purpose

Matomo has some nice displays of analytics that could be imported into the portal analytics page. A few Matomo widgets have been identified to import into the portal analytics page:

- [ ] [Visitors location map](https://matomo-dev.acelab.ca/index.php?module=Widgetize&action=iframe&disableLink=0&widget=1&moduleToWidgetize=UserCountryMap&actionToWidgetize=visitorMap&idSite=2&period=range&date=2021-01-22,today&disableLink=1&widget=1) 
- [ ] [Real-time visitors map](https://matomo-dev.acelab.ca/index.php?module=Widgetize&action=iframe&disableLink=0&widget=1&moduleToWidgetize=UserCountryMap&actionToWidgetize=realtimeMap&idSite=2&period=range&date=last30&disableLink=1&widget=1)
- [ ] [Behaviour transition graph](https://matomo-dev.acelab.ca/index.php?module=Widgetize&action=iframe&disableLink=0&widget=1&moduleToWidgetize=Transitions&actionToWidgetize=getTransitions&idSite=2&period=range&date=last30&disableLink=1&widget=1)

Other widgets that could be added are:
- [ ] Browser information ([engines](https://matomo-dev.acelab.ca/index.php?module=Widgetize&action=iframe&disableLink=0&widget=1&moduleToWidgetize=DevicesDetection&actionToWidgetize=getBrowserEngines&idSite=2&period=range&date=last30&disableLink=1&widget=1
) or [versions](https://matomo-dev.acelab.ca/index.php?module=Widgetize&action=iframe&disableLink=0&widget=1&moduleToWidgetize=DevicesDetection&actionToWidgetize=getBrowserVersions&idSite=2&period=range&date=last30&disableLink=1&widget=1
) available)
- [ ] OS information ([families](https://matomo-dev.acelab.ca/index.php?module=Widgetize&action=iframe&disableLink=0&widget=1&moduleToWidgetize=DevicesDetection&actionToWidgetize=getOsFamilies&idSite=2&period=range&date=last30&disableLink=1&widget=1
) or [versions](https://matomo-dev.acelab.ca/index.php?module=Widgetize&action=iframe&disableLink=0&widget=1&moduleToWidgetize=DevicesDetection&actionToWidgetize=getOsVersions&idSite=2&period=range&date=last30&disableLink=1&widget=1
) available)

## Possible Implementation 

[...]

PR #424, opened 2021-03-15
Analytics updates

liamocn:
## Related issues

#403 #404 #405 #406 #373

## Checklist

- [ ] **DO** Unit tests pass.

## Purpose

This PR updates the portal to include many new analytics features, including dataset and pipeline views, popular keywords and daily visitors.
 
- [ ] Yes
- [ ] No


cmadjar (changes requested):
@liamocn This looks amazing!!!

A few comments:
- the maps, is there a way to have less white space left and right? Or could they be displayed one next to the other?
- transition graph, is there a way to make it display a default transition graph instead of having to select menus multiple times
- the pipeline bubble graph is displaying on top of the graph title on smaller pages
Screen Shot 2021-03-19 at 11 11 46 AM
- number of unique visitors: it was mentioned to have a cumulative histogram. Maybe the bars could be cumulative and there could be a dotted line with the actual number of visits per month displayed on top of it? A bit like this?
Screen Shot 2021-03-19 at 11 19 36 AM
- keywords: I wonder if we should filter out the keywords that are less than 2 characters. At least, the ones with only one character should probably not be displayed.
- pipelines graph: is it possible to map the zenodo ID with the actual pipeline name from the Boutiques descriptors?
- for the number of views, is it possible to add a little eye icon with the number of views in the dataset and pipeline pages?
- when the matomo iframe cannot be displayed, is it possible to either hide the iframe or display a message: temporarily unavailable or something like that?
- the transition graph display is being cut: 
Screen Shot 2021-03-19 at 11 31 41 AM

liamocn:
**the maps, is there a way to have less white space left and right? Or could they be displayed one next to the other?**
They have been moved next to each other. The whitespace at the sides is the same for all pages, so it's a larger task to change that.

**transition graph, is there a way to make it display a default transition graph instead of having to select menus multiple times**
I'm not sure about this. If it's possible it would be in the matomo documentation.

**the pipeline bubble graph is displaying on top of the graph title on smaller pages**
I wasn't able to come up with an acceptable solution for this

**number of unique visitors: it was mentioned to have a cumulative histogram. Maybe the bars could be cumulative and there could be a dotted line with the actual number of visits per month displayed on top of it? A bit like this?**
It's now been changed to a dual-axis chart

**keywords: I wonder if we should filter out the keywords that are less than 2 characters. At least, the ones with only one character should probably not be displayed.**
Done

**pipelines graph: is it possible to map the zenodo ID with the actual pipeline name from the Boutiques descriptors?**
Done

**for the number of views, is it possible to add a little eye icon with the number of views in the dataset and pipeline pages?**
Done

**when the matomo iframe cannot be displayed, is it possible to either hide the iframe or display a message: temporarily unavailable or something like that?**
This is difficult because of how iframes work. By design, the portal won't know what's in the frame. There may be a solution involving callbacks, possibly

the transition graph display is being cut:
Fixed

cmadjar (approved):
@liamocn  This looks truly amazing! Thank you!!!

Issue #418, opened 2021-03-09
[500 error] for McGill EMC: RNA-seq experiment from human brain sample


mathdugre:
## Summary
I was browsing through the portal and noticed that there is currently (Mar. 09 12:36) a **500 error** for the *McGill EMC: RNA-seq experiment from human brain sample* dataset.
I'm not sure if this error is transient or persistent, but I thought it would be good to report it here.

https://portal.conp.ca/dataset?id=projects/mcgill-emc-rna-seq-experiment

cmadjar:
@mathdugre thank you for logging the bug. The PR to fix that issue has been sent on conp-portal (#426)

PR #426, opened 2021-03-16
Fix 500 error for the McGill EMC dataset page

cmadjar:
this fixes the error reported in https://github.com/CONP-PCNO/conp-portal/issues/418.

The `producedBy` field of DATS could be a string, an array or an object. The fix handles the different possibilities when reading the DATS files.


liamocn:
Looks good, thanks!

PR #438, opened 2021-04-01
[analytics] Fix 'null' pipeline being shown in the most viewed pipeline

cmadjar:
An old pipeline has been removed from the portal but showed up in the most viewed pipelines graphs with a label `null`. The change in this PR will ignore the stats for pipelines with no titles.

Before: 
Screen Shot 2021-04-01 at 4 38 56 PM

After: 
Screen Shot 2021-04-01 at 4 41 46 PM

### DIFF
--- a/app/analytics/routes.py
+++ b/app/analytics/routes.py
@@ -160,15 +160,16 @@ def pipelines_views():
                     v.nb_uniq_visitors if v.nb_uniq_visitors is not None else 0)
 
         if not exists and v.label is not None and "/pipeline?id=" in v.label:
-            element = {
-                "url": v.url,
-                "label": v.label,
-                "title": pipelines.get_title_from_id(v.label.split('id=')[1]),
-                "nb_hits": v.nb_hits,
-                "nb_visits": v.nb_visits,
-                "nb_uniq_visitors": v.nb_uniq_visitors if v.nb_uniq_visitors is not None else 0,
-            }
-            elements.append(element)
+            if pipelines.get_title_from_id(v.label.split('id=')[1]):
+                element = {
+                    "url": v.url,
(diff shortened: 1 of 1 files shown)


cmadjar:
@mathdugre @joeyzhou98 @tkkuehn would one of you be willing to review this PR? The review should be straightforward on this one. Thank you!

app/analytics/routes.py
```diff
@@ -160,15 +160,16 @@ def pipelines_views():
...
-            element = {
-                "url": v.url,
-                "label": v.label,
-                "title": pipelines.get_title_from_id(v.label.split('id=')[1]),
-                "nb_hits": v.nb_hits,
-                "nb_visits": v.nb_visits,
-                "nb_uniq_visitors": v.nb_uniq_visitors if v.nb_uniq_visitors is not None else 0,
-            }
-            elements.append(element)
+            if pipelines.get_title_from_id(v.label.split('id=')[1]):
```
joeyzhou98, line 163: You are assuming that `id` is the only query in the url, what if there are multiple queries like `/pipeline?id=123&arg=456` or `/pipeline?arg=123&id=456&arg2=789`. Would this scenario be possible?

joeyzhou98 (changes requested):

app/analytics/routes.py
cmadjar: @joeyzhou98 I just looked into the database just in case. The URLs gathered by Matomo and present in the database table are always of the form:
`https://portal.conp.ca/pipeline?id=zenodo.1234567`
There is no other argument in the URL of a tool on the portal.

joeyzhou98 (approved):

PR #441, opened 2021-04-14
Add the greyed out button to not downloadable datasets with explanation

cmadjar:
This adds a layout for the non-downloadable datasets with an explanation as to why the dataset cannot be downloaded.

- Layout for a dataset that can be downloaded:
Screen Shot 2021-04-14 at 10 58 37 AM

- Layout for a dataset that cannot be downloaded:
Screen Shot 2021-04-14 at 10 57 30 AM

### DIFF
--- a/app/templates/dataset.html
+++ b/app/templates/dataset.html
@@ -228,26 +228,41 @@
 
 <div class="d-flex flex-column p-2">
   <div class="d-flex flex-row align-items-center">
-    <a name="downloadInstructions">
-    </a>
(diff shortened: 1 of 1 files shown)


xlecours (approved):
I you are confident in these changes being the same than the one we tested on dev, it is a go :)

cmadjar:
@xlecours I just pulled it on portal-dev to test it out for a sanity check and it has the expected behaviour. Should be good to go on prod! Thank you!

PR #446, opened 2021-04-22
Add https:// in front of matomo URL for the update_analytics function of flask

cmadjar:
When the analytics got pulled to the portal, the config was changed from `base_url` to `https://base_url` causing the JS tracker to fail. 

In this PR, we modify the `update_analytics` cli function to add `https://` in front of the Matomo URL and add a note next to the config on the `flaskenv.template` to avoid any confusion in the future.

There are probably more elegant ways to deal with this but we need a short term fix so that the updates to the analytics can run and the tracking code can work correctly.


cmadjar:
@emmetaobrien @joeyzhou98 if you guys have some time to take a look. I tested locally and that works well

Issue #447, opened 2021-04-22
Add analytics on the downloaded datasets via the button on portal


cmadjar:
## Purpose

The same way as we have the "number of views" for datasets, add the number of downloads information based on what Matomo is recording.

TODOs:
- [x] create a table that will host the daily dataset download stats (similar to the daily dataset view stats) https://github.com/CONP-PCNO/conp-portal/pull/448
- [x] query the Matomo API to populate the created table https://github.com/CONP-PCNO/conp-portal/pull/448
- [x] create the routes to gather the numbers from the SQL table
- [x] show the number of downloads in the dataset card next to the number of views (below the logo of the dataset)
- [ ] create a summary graph to display on the analytics page

cmadjar:
Query to check the stats on the DB: 
```
SELECT label, SUM(nb_hits) FROM matomo_daily_portal_download_summary WHERE label LIKE "/data/%" GROUP BY label;
```

Results on 2021-08-05:
![Screen Shot 2021-08-05 at 5 15 11 PM](https://user-images.githubusercontent.com/1402456/128422215-bad2ec9c-0883-414b-ae44-d72a485f3780.png)

PR #448, opened 2021-04-22
Query Matomo for download button statistics

cmadjar:
### Description

This PR adds the analytics summary for the downloads to a new Matomo SQL table in the portal database.

Structure of the table `matomo_daily_dataset_portal_download_summary`:
```
    - date             = date of the requested summary statistics (day)
    - url              = page URL
    - label            = page label
    - nb_hits          = number of views on this page
    - nb_visits        = number of visits (30 min of inactivity considered a new visit)
    - nb_uniq_visitors = number of unique visitors
    - sum_time_spent   = total time spent on this page, in seconds
    - segment          = segment with download URL
```

The Matomo API endpoint queried is the Actions.getDownloads method. The result of the following query is inserted into the `matomo_daily_dataset_portal_download_summary`

### Related to 

Part 1 and 2 of #447


cmadjar:
Notes to self for tomorrow: rename the table and the python functions as that table will contain download information for all downloadable items not just the dataset items.

cmadjar:
@joeyzhou98 @mathdugre @xlecours @tkkuehn: ready for review! :)

Note: this is not adding any visualizations. Just querying the Matomo analytic instance of the portal and insert daily stats on the downloads into a new database table that will be displayed in a separate PR

app/models.py
```diff
@@ -353,6 +353,36 @@ def __repr__(self):
...
+    - nb_hits          = number of views on this page
+    - nb_visits        = number of visits (30 min of inactivity considered a new visit)
+    - nb_uniq_visitors = number of unique visitors
+    - sum_time_spent   = total time spent on this page, in seconds
+    - segment          = segment with download URL
+    """
+    __tablename__ = 'matomo_daily_portal_download_summary'
+
+    id = db.Column(db.Integer, primary_key=True, autoincrement=True)
+    date = db.Column(db.String(12))
```
joeyzhou98, line 373: Have you considered the db.Date type here? https://docs.sqlalchemy.org/en/13/core/type_basics.html#sqlalchemy.types.Date

app/models.py
```diff
@@ -353,6 +353,36 @@ def __repr__(self):
...
+    url = db.Column(db.Text)
+    label = db.Column(db.String(256))
```
joeyzhou98, line 375: why is url of type Text and not label or segment?

joeyzhou98 (commented):
Just general curiosity questions

app/models.py
cmadjar: yes. It was the first thing I did when I created the Matomo tables on another PR but the `flask db upgrade` command failed when the date field was db.Date so I had to change it to string unfortunately. I don't remember the exact details of the failure but it was the only solution to fix the failure...

app/models.py
cmadjar: the URL could be longer than 256 characters, which caused the insert statements to fail sometimes so I had to modify the type to Text instead of String. That was not an issue for the label or segment fields

joeyzhou98 (approved):

PR #454, opened 2021-04-28
Fix graph of cumulative number of datasets and pipelines

cmadjar:
This fixes two bugs in the "Cumulative number of datasets and pipelines" that resulted in the wrong number being shown for the number of datasets in April 2021 and the number of datasets in November 2020 and December 2020. 

See comments in the PR for bug fix details.

## Related issues

#453


app/static/lib/conp-react/src/charts/TotalDatasetsPipelines/index.js
```diff
@@ -152,35 +154,30 @@ const TotalDatasetsPipelines = (props) => {
         };
 
         chartData.datasets.elements.forEach(element => {
-            const dateAdded = new Date(element.dateAdded);
+            const dateAdded = new Date(element.dateAdded.concat('T00:00:00'));
```
cmadjar, line 157: Because the format of the `dateAdded` field of the JSON response is of format `YYY-MM-DD`, the date is parsed in UTC (GMT) which resulted in returning a date for the day before.

See: https://stackoverflow.com/questions/7556591/is-the-javascript-date-object-always-one-day-off

Adding the `T00:00:00` will ensure that the date object created correspond to the correct day.

app/static/lib/conp-react/src/charts/TotalDatasetsPipelines/index.js
```diff
@@ -152,35 +154,30 @@ const TotalDatasetsPipelines = (props) => {
...
             }
 
-            if (!axes.datasets[dateAdded.getFullYear()][dateAdded.getMonth() + 1]) {
-                axes.datasets[dateAdded.getFullYear()][dateAdded.getMonth() + 1] = 1;
+            if (!axes.datasets[dateAddedYear][dateAddedMonth]) {
+                axes.datasets[dateAddedYear][dateAddedMonth] = 1;
             }
             else {
-                axes.datasets[dateAdded.getFullYear()][dateAdded.getMonth() + 1] += 1
+                axes.datasets[dateAddedYear][dateAddedMonth] += 1
```
cmadjar, line 169: These changes are just for clarity of the code.

app/static/lib/conp-react/src/charts/TotalDatasetsPipelines/index.js
```diff
@@ -152,35 +154,30 @@ const TotalDatasetsPipelines = (props) => {
...
-
         const today = new Date();
 
-        if (!Object.keys(axes.datasets).includes(today.getFullYear())) {
+        if (!Object.keys(axes.datasets).includes(today.getFullYear().toString())) {
             axes.datasets[today.getFullYear()] = {}
         }
 
-        if (!Object.keys(axes.pipelines).includes(today.getFullYear())) {
+        if (!Object.keys(axes.pipelines).includes(today.getFullYear().toString())) {
```
cmadjar, line 180: `today.getFullYear()` is of type `number` but the keys of an object are of type `string` so need to convert in the comparison here. If we do not do any conversion to string, then all the numbers of additional dataset for the current year will be 0, which is why the number of datasets in April 2021 was 49 and not 50.

cmadjar:
@glatard that will fix the issue in the "cumulative number of datasets and pipelines" graph that you mentioned during the meeting.

It only took me the whole day to finally fix it. React is definitely not my forte ;) 

While trying to figure out the issue of April, I found another bug which is resolved in this PR as well.

cmadjar:
@joeyzhou98 @tkkuehn @mathdugre if one of you feels like reviewing this. It has been tested locally on my computer in different browsers.

joeyzhou98 (approved):
LGTM

Issue #462, opened 2021-05-12
Small rewording on the dataset detailed page


cmadjar:
- Replace "Download This Version" download button text with "Download current version"

- Clarify contrast between direct download and datalad sections by changing section title from "DataLad Dataset Download Instructions" to "Download using DataLad"

PR #463, opened 2021-05-13
Rewording of download button and datalad download section header on detailed dataset page

cmadjar:
This replaces the text of the download button and Datalad instruction sections:

 
- Replaced "Download This Version" download button text with "Download current version"
 
- Clarified the contrast between direct download and datalad sections by changing section title from "DataLad Dataset Download Instructions" to "Download using DataLad"

#462


cmadjar:
@emmetaobrien when you have some time, could you review this? Thanks!

PR #465, opened 2021-05-17
Process CONP datasets in CBRAIN , Issue 348

mandana-mazaheri:
---
Title: Change the style of CBrain button
title: 'Feature implementation for issue #<348>'
labels: enhancement
assignees: ''Mandana Mazaheri"

---

## Checklist

- [ ] **DO** Unit tests pass.

## Purpose
<To reduce the size of CBrain button and apply dynamic resizing for that >

## Current behavior
<The CBrain button was too large and was in the same division as the next column so whenever the size of the window changed these two columns would merge or go out of the border>

## New behavior
<The CBrain button is smaller and its size is automatically compatible with page size>
 
- [ ] Yes
- [ ] No

### DIFF
--- a/app/static/lib/conp-react/src/DatasetElement/index.js
+++ b/app/static/lib/conp-react/src/DatasetElement/index.js
@@ -208,7 +208,6 @@ const DatasetElement = props => {
               </div> : null}
           </div>
         </div>
-        
         <div className="col col-lg-2 d-flex flex-column justify-content-top align-items-center p-2">
         <div class="row align-items-top width-auto">
         
@@ -226,8 +225,6 @@ const DatasetElement = props => {
 			src="static/img/cbrain-icon-grey.png" style={{maxWidth: '60px'}}/>
 	    </a>}
         </div>
-        
-        
         <div className="col col-lg-8 d-flex flex-column justify-content-top align-items-center p-2"  >
           <h7>DOWNLOAD</h7>
           <div className="d-flex flex-column">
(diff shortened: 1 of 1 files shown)


cmadjar (changes requested):
@mandana-mazaheri see attached comment to modify the CBRAIN button so that it has a fixed size when reducing the window.

For the padding, you will need to modify line 109 to the following:
```
      <div className="row pr-4">
```
(addition of the px-4 bootstrap to ensure there is padding on the left and right side of the card which will avoid the buttons to overlap with the card border.)

Finally, maybe we can set a time together where we go through your npm install because the builds need to be generated for your PR.

cmadjar:
@mandana-mazaheri have you had a chance to work on this?

mandana-mazaheri:
Hi Cecile, thanks for following up.

Yes I applied what you mentioned before but the button size changed to the
larger size than the current one. I attached the photo of that. I tried
different px sizes but didn’t change anything.

I’m working more on the dashboard of pipeline executions now. But I’ll try
to figure why it happens to the button here.

mandana-mazaheri:
I’m working on another laptop (the linux one) for the portal and that’s why
I used my phone. 😁

mandana-mazaheri:
To reduce the size of CBRAIN button and apply the requested changes

cmadjar:
@mandana-mazaheri thank you for the changes. Could you please generate the npm build files and add them to the PR?

```
npm install
npm run build
```

cmadjar:
Closing in favour of #471 as discussed with @mandana-mazaheri offline.

Issue #412, opened 2021-02-22
Add prepopulated queries to the Nexus search


cmadjar:
## Purpose

Create and prepopulate a few Nexus queries in the Advance search page so that users can see examples of Nexus queries or simply use prepopulated queries.

## Related issues 

A follow up to the work that has been done in #276 and #377

cmadjar:
@zxenia feel free to add more notes relevant to this issue. Thanks!

cmadjar:
@zxenia any updates on this? I forgot. If you could update the issue to keep track on where things are at, that would be awesome. Thank you!

zxenia:
Hi @cmadjar, it's coming along.
Sebastian created some nice queries, we planned to use them.
Meanwhile, I started to look into how to add it to the widget I used for sparql editor.

PR #481, opened 2021-08-27
Add example SPARQL queries

zxenia:
## Purpose
This PR adds five predefined example SPARQL queries.
Each query prepopulates a separate tab in the SPARQL editor at `/sparql`.

## Related issues
#412

## Current behaviour
Only one tab with a generic query to query for 10 random triples.

## New behaviour
Overall six tabs in the query editor: one default active tab to start the query and five tabs with example queries that follow our schema.org context mappings.

- [ ] Yes
- [x] No

example_sparql_queries

### DIFF
--- a/app/search/queries.py
+++ b/app/search/queries.py
@@ -7,7 +7,7 @@ SELECT DISTINCT ?title ?license_name WHERE {
 ?dataset a sdo:Dataset;
   sdo:name ?title;
   sdo:license ?license.
-  ?license sdo:name ?license_name.
+?license sdo:name ?license_name.
 FILTER (?license_name = "CC BY-NC-SA"^^sdo:Text)
 }
 """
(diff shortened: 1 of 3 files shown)


cmadjar:
@zxenia why is PREVENT-AD registered dataset duplicated 7 times in the query results that display it? Is there something wrong with the Nexus data?

Other than that, looks all good to me.

zxenia:
@cmadjar I looked into it and yes, currently the data in Nexus has several prevent _PREVENT-AD registered data_ datasets - each of them has a distinct unique identifier, that's why they are treated as different datasets. Perhaps something went wrong with the data ingestion therefore this dataset is duplicated and assigned a new id. Sebastian and I planned to update the data in Nexus soon (including the addition of extended context) and coordinate it with Adeel, so we will look into this issue.
If everything else is ok I'd prefer it to be merged since this PR only takes care of embedding the queries in the UI. We can have another issue with the Nexus data update. Thank you!

cmadjar:
@zxenia OK. Perfect. Could you create an issue and assign it to you and Sebastian regarding the Nexus update and duplication of some datasets please? Thanks!

Issue #484, opened 2021-09-08
Update Nexus data


zxenia:
## Purpose
The data in Nexus needs to be updated since the context mapping to schema.org was improved.
Also currently there is possible data duplication in Nexus for PREVENT-AD registered data dataset - it needs to be investigated.

Ideally, this update will also include migration to proper dataset identifiers: instead of UUID to a stable identifier from the CONP database. This improvement will allow being redirected from Nexus search results to an individual dataset page. Currently, Nexus results contain dataset resolvable ids that redirect to Nexus dataset representation in JSON.

surchs:
@zxenia what's the desired format for the CONP database URIs for datasets? 

At the moment I can see portal datasets have a URI of the form:
`https://portal.conp.ca/dataset?id=projects/PERFORM_Dataset__one_control_subject`

How are these generated (e.g. when we have new data come in), and do we want to use this form as the resource ID? Something like just 
`https://portal.conp.ca/dataset/PERFORM_Dataset__one_control_subject`
might be cleaner, but I am not sure if there is a reason not to adopt this.

Adopting these dataset based resource IDs is going to be helpful also for updating the existing records in nexus when we make changes to the meta-data on our end - that is, as opposed to just deprecating the original version and creating a new record. So it would help to make sure that we also have an easy way of tying the resource ID / URI unambiguously to a given dataset in the future.

zxenia:
Hi @surchs 

I think we will need to stick with this URIs `https://portal.conp.ca/dataset?id=projects/PERFORM_Dataset__one_control_subject`
because it's how the conp portal route for a single dataset is designed. If we have those URIs in Nexus then the search results will be immediately resolvable to a dataset's detail page.
To have this `https://portal.conp.ca/dataset/PERFORM_Dataset__one_control_subject` will probably require changes to url path.
And if we change them then we have to make the old urls resolvable to those new ones.

> How are these generated

First, I thought that a dataset's database id (or path)  is generated from its title and prefixes with `projects/` - the title is being split by words and added dashes or underscores.
But I cross-checked and I see that it's not true and seems like it's arbitrary looking at the below examples:

```
title: BigBrain dataset - 3D Classified Volumes (derived dataset)
url: /dataset?id=projects/BigBrain_3DClassifiedVolumes

title: Multicenter Single Subject Human MRI Phantom
url: /dataset?id=projects/multicenter-phantom
```

@cmadjar Do you know how a dataset's id is generated in the database?

[...]

PR #512, opened 2021-12-22
Updated canned queries for new nexus context

### LINKED ISSUE #506: Update SPARQL canned queries to use the new dataset-url attribute
We are preparing to use the updated data in nexus (#484 ). 
As part of this update, Adeel (Nexus) will introduce a new attribute at the dataset level to contain the CONP-URL.
The goal is that Nexus can continue handling the dataset UUIDs, but that we also have a way to show our users the URL of the dataset rather than some less useful reservoir URL.

To make use of the new attribute we should:
- [x] pick an attribute name that we can then ask Adeel to add in his workflow
- [x] change our current set of [canned queries](https://github.com/CONP-PCNO/conp-portal/blob/master/app/search/queries.py) to display the value of this attribute, rather than the dataset UUID as it does currently

surchs:
## Related issues
Fixes #506 
#484 
https://github.com/CONP-PCNO/conp-dataset/issues/595

## Purpose
Nexus has rolled out the new schema.org contexts and also added a new attribute to each dataset to link to the corresponding conp portal website. The updated canned queries address these changes.

## Current behaviour
Queries are returning deprecated datasets and are not making use of the new conp portal links.

## New behaviour
canned queries
- will only return current datasets (undeprecated)
- show the link to the conp dataportal website for each result dataset
- make use of new schema.org terms that were previously custom conp/DATS terms
- have simplified syntax in some places
 
- [ ] Yes
- [X] No

### DIFF
--- a/app/search/queries.py
+++ b/app/search/queries.py
@@ -41,8 +41,8 @@ SELECT DISTINCT ?dataset ?title ?format WHERE {
   sdo:name ?title;
   nexus:deprecated false;
   sdo:distribution ?distribution.
-?distribution conp:formats ?format.
-FILTER (regex(?format, "MINC", "i"))
+?distribution sdo:encodingFormat ?format.
+FILTER (regex(str(?format), "MINC", "i"))
 }
 """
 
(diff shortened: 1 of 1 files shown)


surchs:
Any datasets that are not currently passing the validator will also not be converted to the nexus graph and will therefore not be shown in the SPARQL queries - they still show up via regular search though. See CONP-PCNO/conp-dataset#595

cmadjar (approved):

Issue #349, opened 2020-10-07
Process CONP datasets in CBRAIN using CONP pipelines


cmadjar:
## Tasks

- In the CONP portal, create an interface to select a pipeline from a dataset, and/or to select a dataset from a pipeline
- From this interface, redirect to a pre-populated CBRAIN launch form

## How

Needs discussion, it might be a bit tricky, as fine-grained file selection in the dataset might be necessary.

## Who

CONP portal developers: @liamocn, @xlecours

## Related issues (optional)

#63 

Depending on #347 and #348

glatard:
It should be doable through the CBRAIN API. @natacha-beck did something similar.

natacha-beck:
Since all CONP dataset that is currently in CBRAIN is setup as one DataProvider. The API should select the file present on the DataProvider (via the API call), then launch a Tool on this data still via the API.

tkkuehn:
I'll be away for the roadmap meeting next Wednesday, so a quick update on this: I have a first draft of the UI portion working and and I'm working on figuring out the best series of API calls to properly launch a task. This should be done by the end of February. Sorry for the delay!

PR #524, opened 2022-03-02
Run CONP pipelines on CONP datasets using CBRAIN

tkkuehn:
---
name: Run CONP pipelines on CONP datasets using CBRAIN
about: Propose modifications in the code
title: 'Feature implementation for issue #349'
labels: enhancement
assignees: ''

---

## Related issues
Addresses #349.

## Checklist

- [ x] **DO** Unit tests pass.

## Purpose
This PR adds the capability to open a pair of one CONP dataset and one CONP pipeline in CBRAIN, presenting the user with the CBRAIN page to configure the pipeline with that dataset.

## Current behaviour
Users must either pick a pipeline or a dataset, then find the other pipeline/dataset manually on CBRAIN.

## New behaviour
A dropdown menu is shown below the CBRAIN button for each pipeline or dataset in which an entity of the other type can be selected. If a pair are selected, clicking the CBRAIN button will open the dataset in a new window to load it, and a modal window on the portal. The intention is for the user to log in and load the dataset in the CBRAIN window, then closing the modal on the CONP Portal will open the pipeline page with the dataset pre-loaded.
 
- [ x] Yes
- [ ] No

## Implementation Detail
This workflow is a little awkward (the user still needs to a fair amount of clicking and moving between windows), but I couldn't think of a better way to do it subject to the constraints imposed by CBRAIN (i.e. the dataset page needs to be opened followed by the pipeline page, and CBRAIN cannot be loaded in an iframe). I am certainly open to implementing this in a different way if there's a better feasible user experience that I haven't thought of! It may be best to go over it on the next dev call, but here it is to take a look at.


cmadjar (changes requested):
@tkkuehn Thank you for working on this!! 

This does not seem to be working for me. See below for suggestions. Thank you :-)


# Dataset page

- [ ] When I select a pipeline to be run on a specific dataset, nothing happens.
Screen Shot 2022-03-08 at 3 43 30 PM
s

- [ ] If I then click on the CBRAIN button, I get redirected to the dataset installed on CBRAIN but no tool preselected. Then I go back to my sandbox portal and see this screen:
Screen Shot 2022-03-08 at 3 44 57 PM


# Pipeline page

- [ ] clicking on the CBRAIN button no longer open a new tab with the tool preselected on CBRAIN - I think this happens when no datasets are selected. We should still allow for opening the tool in CBRAIN without selecting datasets. People might want to run a tool on their own dataset

- [ ] Same comment as for the dataset, when I select a dataset, nothing happens.  

# General comments

Would it be possible to improve the user interface to make it more intuitive? 
- [ ] Once the tool is selected, have the pop up show with some text "You are about to go to CBRAIN to run tool X on dataset Y" and then a button saying "Go to CBRAIN" or something like that?
- [ ] The select drop down should be in the same kind of format as the ones used by the filters for consistency.

tkkuehn:
Thanks for your review @cmadjar, I've addressed your comments (including a couple of bugs, sorry about that!).

To clarify, the expected workflow now is:

1. Select a pipeline/dataset from the dropdown (this defaults to "None", and changing it does not launch CBRAIN, you still need to press the CBRAIN button).
2. Press the CBRAIN button (if no pipeline/dataset is selected from the dropdown, this will just launch CBRAIN, no change from the current UX).
3. The popup explains which dataset and pipeline have been selected, and prompts the user to use the buttons to open the dataset and pipeline page in that order.
4. Once the pipeline has been launched, the popup closes.

I wasn't 100% sure whether you wanted to have the popup launch immediately when a pipeline/dataset is chosen from the dropdown, or when a user presses the CBRAIN button. I can set it up either way, but it's currently the latter (i.e. the user needs to press the CBRAIN button to open the popup).

tkkuehn:
Hi @cmadjar, thank you for the review.

I can't reproduce that bug locally -- When you see this bug, are the correct dataset and pipeline names shown in the pop up? If not, there's certainly something wrong that I'm not reproducing for some reason.

If you do see the correct dataset and pipeline names, it's possibly more of a lack of clarity in the pop-up text. To get the correct dataset loaded, the user always needs to first press the grey "Load dataset" button, then press the blue "Prepare tool" button. If a user only presses the "Prepare tool" button after selecting a dataset from the dropdown on the pipeline page, the dataset will not be loaded. If this is the case, we could disable the "Prepare tool" button until the "Load dataset" button is pressed to make the required steps a bit clearer.

In any case, I've added the requested note about compatibility between CBRAIN pipelines and datasets (And clarified that you need to be logged into CBRAIN first for this to work).

tkkuehn:
@cmadjar I made the changes Tristan suggested (moving the pipeline/dataset selection into the pop-up and only showing one instruction at a time), so this could use another review.

I also accidentally ran a code formatter on the dataset and pipeline element react files, so those files have a lot of formatting changes that don't affect their function. Sorry! I can go back and revert the formatting if you want.

prioux:
Hi everyone. I just watched the video of the Zoom session where this was presented this morning. Bryan provided me a link to this PR here, I think it's an appropriate place for my comments.

As indicated in the description of the PR, the workflow is cumbersome. The reason for this is that within the CBRAIN interface we have only had to provide users the ability to select files and select tools as separate steps. In that context it works fine:

1) CBRAIN has direct links to switch the state of the file manager to filter by files from a particular project;
2) Way back in Oct 2020 Natacha implemented an interface hack so that CONP could link directly to the file manager page with a tool already 'prepared'. Note that in CBRAIN we don't use that at all, it's entirely for you.

So the workflow described here requires invoking each one separately. One CBRAIN request with `/groups/switch/NNN` and one request with `/userfiles?prepare_tool_id=MMM`.

But it doesn't have to be like that. Why didn't you simply ask us to create a unified request that does both? It would have been simple on our side, you know. Something like `/userfiles?group_switch=NNN&prepare_tool_id=MMM`. On your side you could provide both values, or just one of them.

Let me know if you guys are interested.

tkkuehn:
Hi @prioux,

Thank you for taking a look at the Zoom session for this!

Frankly, I just didn't think to ask the CBRAIN team to implement a unified request to prepare a dataset and tool together. If it's reasonably simple to do, it would make the implementation on our side easier and the user experience smoother, so (in my opinion) we would certainly be interested in the unified request.

Thanks again for taking the time to look at it, and I'll certainly be more proactive about reaching out if I run into a feature request for CBRAIN like this that would make a CONP feature easier to put together.

cmadjar:
That would be wonderful to have the dataset ID and pipeline ID in a CBRAIN URL that could redirect to the loaded dataset + tool. 

@prioux How long do you think it would take to implement this on the CBRAIN side?

prioux:
2 hours for me. 5 to 20 days for anyone else.

I think I have free time late in September, 2023.

prioux:
I was joking btw. Basically, ask Bryan to schedule some time for this and I'll do it.

cmadjar:
Added the blocked label as more development will happen once CBRAIN has a request where we could specify both the dataset and the pipeline to load at once.

prioux:
Good news I've implemented the CBRAIN side.

Generate your URLs to CBRAIN with "switch_group_id=NNN" in the query parameters.

prioux:
The CBRAIN side code is currently under review:

https://github.com/aces/cbrain/pull/1235

tkkuehn:
@cmadjar I've updated the PR to use the new CBRAIN URL scheme, so it's ready for a another look but won't work until that change is deployed to CBRAIN.

app/static/lib/conp-react/src/CbrainModalDataset/index.js
```diff
@@ -0,0 +1,76 @@
...
+    $("#cbrainModal").modal("handleUpdate");
+    return;
+  });
+
+  return ReactDOM.createPortal(
+    <div>
+      <p>
+        You're about to load the dataset <strong>{title}</strong> on
+        CBRAIN. Please select a pipeline to process this dataset (or "None" to
+        load only the dataset).
```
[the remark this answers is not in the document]
cmadjar: Since several datasets actually have the word dataset in their title, it can read funny.

app/templates/common/base_main.html
```diff
@@ -820,6 +820,25 @@ <h5>Create or Edit a DATS.json File Tab</h5>
...
+  id="cbrainModal"
+  tabindex="-1"
+  role="dialog"
+  aria-labelledby="modalCenterTitle"
+  aria-hidden="true"
+>
+  <div class="modal-dialog modal-dialog-centered modal-lg" role="document">
+    <div class="modal-content">
+      <div class="modal-header">
+        <h4 class="modal-title w-100 text-center" id="modalLongTitle">CBRAIN Dataset Select and Pipeline Launch</h4>
```
cmadjar, line 834:
```suggestion
        <h4 class="modal-title w-100 text-center" id="modalLongTitle">CBRAIN Launch</h4>
```

cmadjar: I wonder if the title of the modal window should not be shorter

app/static/lib/conp-react/src/CbrainModalPipeline/index.js
```diff
@@ -0,0 +1,69 @@
...
+    return;
+  });
+  return ReactDOM.createPortal(
+    <div>
+      <p>
+        You're about to load the pipeline <strong>{title}</strong> on
+        CBRAIN. Please select a dataset to process with this pipeline (or "None"
+        to load only the pipeline).
+      </p>
+      <p>Note that not all pipelines are compatible with all datasets.</p>
```
cmadjar, line 23: This sentence is only present for the pipeline modal, not in the dataset modal. I guess it should be represented on both?

app/static/lib/conp-react/src/CbrainModalPipeline/index.js
```diff
@@ -0,0 +1,69 @@
...
+  const getDatasetId = (datasetUrl) =>
+    `switch_group_id=${datasetUrl.split("?id=")[1]}`;
+  useEffect(() => {
+    $("#cbrainModal").modal("handleUpdate");
+    return;
+  });
+  return ReactDOM.createPortal(
+    <div>
+      <p>
+        You're about to load the pipeline <strong>{title}</strong> on
```
cmadjar, line 19:
```suggestion
        You're about to launch <strong>{title}</strong> on
```

cmadjar: same comment as for the dataset but used the word launch instead of load (load for dataset and launch for pipeline seemed logical to me but does not mean it is necessarily)

requirements.txt
```diff
@@ -12,7 +12,7 @@ blinker==1.4
 boto==2.49.0
 boutiques==0.5.25
 certifi==2018.11.29
-cffi==1.12.3
+cffi==1.15.0
```
cmadjar, line 15: are those requirements associated to your PR? If not, maybe this could be a separate PR to update the requirements?

app/static/lib/conp-react/src/PipelineElement/index.js
```diff
@@ -1,95 +1,163 @@
...
 import ViewsIcon from "../social/ViewsIcon"
 import DownloadsIcon from "../social/DownloadsIcon"
 import ArkIdElement from "../ArkIdElement"
+import CbrainModalPipeline from "../CbrainModalPipeline";
 
-const PipelineElement = props => {
+const PipelineElement = (props) => {
   const { authorized, ...element } = props;
+
+  const [modalOpen, setModalOpen] = useState(false);
```
cmadjar, line 12:
```suggestion
  const [cbrainModalOpen, setCbrainModalOpen] = useState(false);
```

app/static/lib/conp-react/src/PipelineElement/index.js
```diff
@@ -1,95 +1,163 @@
...
+
+  const openModal = (datasetTitle, datasetUrl, pipelineTitle, pipelineUrl) => {
```
cmadjar, line 14:
```suggestion
  const openCbrainModal = (datasetTitle, datasetUrl, pipelineTitle, pipelineUrl) => {
```

app/static/lib/conp-react/src/PipelineElement/index.js
```diff
@@ -1,95 +1,163 @@
...
+    $("#cbrainModal").modal("show");
+    setModalOpen(true);
```
cmadjar, line 16:
```suggestion
    setCbrainModalOpen(true);
```

app/static/lib/conp-react/src/PipelineElement/index.js
```diff
@@ -1,95 +1,163 @@
...
+  };
+  $("#cbrainModal").on("hidden.bs.modal", (event) => setModalOpen(false));
```
cmadjar, line 18:
```suggestion
  $("#cbrainModal").on("hidden.bs.modal", (event) => setCbrainModalOpen(false));
```

app/static/lib/conp-react/src/PipelineElement/index.js
cmadjar: just because there will be other modal windows on the page so better to name it with CBRAIN for transparency

app/static/lib/conp-react/src/PipelineElement/index.js
```diff
@@ -1,95 +1,163 @@
...
+    item.uri ? (
+      <div className="col">
+        <span
+          key={key}
+          data-toggle="tooltip"
+          title="Run Pipeline"
+          style={{ maxWidth: "140px" }}
+        >
+          {
+            <button className="btn" onClick={openModal}>
```
cmadjar, line 30:
```suggestion
            <button className="btn" onClick={openCbrainModal}>
```

app/static/lib/conp-react/src/DatasetElement/index.js
```diff
@@ -1,253 +1,297 @@
...
-  const [datasetErrorState, setDatasetErrorState] = useState(false)
+  const [metadataErrorState, setMetadataErrorState] = useState(false);
+  const [datasetErrorState, setDatasetErrorState] = useState(false);
 
-  const [metadataErrorText, setMetadataErrorText] = useState("")
-  const [datasetErrorText, setDatasetErrorText] = useState("")
+  const [metadataErrorText, setMetadataErrorText] = useState("");
+  const [datasetErrorText, setDatasetErrorText] = useState("");
+
+  const [modalOpen, setModalOpen] = useState(false);
```
cmadjar, line 25:
```suggestion
  const [modalCbrainOpen, setCbrainModalOpen] = useState(false);
```

app/static/lib/conp-react/src/DatasetElement/index.js
```diff
@@ -1,253 +1,297 @@
...
-      }
-      );
-  }
+        setDatasetSpinnerState(false);
+      });
+  };
+
+  const openModal = () => {
+    $("#cbrainModal").modal("show");
+    setModalOpen(true);
```
cmadjar, line 130:
```suggestion
    setCbrainModalOpen(true);
```

app/static/lib/conp-react/src/DatasetElement/index.js
```diff
@@ -1,253 +1,297 @@
...
+  };
+  $("#cbrainModal").on("hidden.bs.modal", (event) => setModalOpen(false))
```
cmadjar, line 132:
```suggestion
  $("#cbrainModal").on("hidden.bs.modal", (event) => setCbrainModalOpen(false))
```

app/static/lib/conp-react/src/DatasetElement/index.js
```diff
@@ -1,253 +1,297 @@
...
-                    <span className="sr-only">Loading...</span>
+          </div>
+          <div className="col col-lg-3 d-flex flex-column justify-content-top align-items-center pr-2">
+            <div className="row align-items-top width-auto">
+
+              <div className="col col-lg-6 d-flex flex-column justify-content-top align-items-center p-2 pr-4">
+                <h7><strong>PROCESS</strong></h7>
+                  {element.cbrain_id ? (
+                    <>
+                      <button className="btn" onClick={openModal}>
```
cmadjar, line 235:
```suggestion
                      <button className="btn" onClick={openCbrainModal}>
```

app/static/lib/conp-react/src/DatasetElement/index.js
```diff
@@ -1,253 +1,297 @@
...
+                  {authIcons.map((icon, index) => <div key={"authIcon_" + index}
+                                                       className="text-center p-1">{icon}</div>)}
+                </div>
               </div>
             </div>
           </div>
         </div>
       </div>
-    </div>
+      {modalOpen ? (
```
cmadjar, line 287:
```suggestion
      {cbrainModalOpen ? (
```

app/static/lib/conp-react/src/PipelineElement/index.js
```diff
@@ -1,95 +1,163 @@
...
-        <div className="d-flex justify-content-end align-items-center">
-          {platforms}
+        <div className="col-sm-12 col-md-3 d-flex align-items-center justify-content-end">
+          <div className="d-flex justify-content-end align-items-center">
+            {platforms}
+          </div>
         </div>
       </div>
-    </div>
+      {modalOpen ? (
```
cmadjar, line 153:
```suggestion
      {CbrainModalOpen ? (
```

cmadjar (changes requested):
@tkkuehn Thank you for updating the PR!!

I added some comments in the code. In summary:
- relabel some of the `modalOpen` variable to `cbrainModalOpen` etc in `DatasetElement/index.js` and `PipelineElement/index.js` since other modal windows will be opened from the card (especially for the dataset download)
- modified slightly the text in the modal window

I find the table of pipelines a little intense and big on the screen and I have a feeling others might think that too based on the feedback I had last week on the card redesign. I wonder if we could do something like that instead:
Screen Shot 2022-05-06 at 2 59 56 PM
Note: don't change anything on the layout yet. Might be best to discuss that on Wednesday after a demo of the changes. Would that make sense?

prioux:
I've searched the diffs of this pull request, and I could find the string 'switch_group_id' but not 'prepare_tool_id'. Is this normal? That seems weird given your website needs to create URLs with both.

tkkuehn:
Great, thanks! A few arbitrarily selected tests indicate that this is working properly.

> I've searched the diffs of this pull request, and I could find the string 'switch_group_id' but not 'prepare_tool_id'. Is this normal? That seems weird given your website needs to create URLs with both.

This is normal -- The URLs to prepare tools for this PR ultimately come from [here](https://github.com/CONP-PCNO/conp-portal/blob/master/app/static/pipelines/cbrain-conp-pipeline.json) and the code in this PR gets the `prepare_tool_id` part by slicing those URLs.

cmadjar (approved):
tested. LGTM 👍

PR #527, opened 2022-03-08
Query GitHub analytics for number of clones

cmadjar:
The changes in this PR adds two new analytics table to store GitHub analytics queried via the GitHub API on the different datasets present in the portal:
- github_daily_views_count -> stores the number of views for a given date and repository (a.k.a. dataset) on GitHub
- github_daily_clones_count -> stores the number of clones for a given date and repository (a.k.a. dataset) on GitHub (can serve as a DataLad download count estimation)

Those two tables will be updated with the cronjob that is running to update the Matomo analytics as well (`flask update_analytics`)

**Major confounding factors in the numbers:**
- the daily tests run on conp-dataset by Circle CI would create clones (which explains why there is always at least 1 unique clone counted per dataset every day...)
- the archiver will also clone a dataset every time it is updated (including DATS and README changes)
- CONP developers also contribute to the number of clones when testing, updating or downloading data...


cmadjar:
This has been thoroughly tested on both my sandbox and on portal-dev.conp.ca. I think I am the only person with a Matomo account so not many people can really test the `flask update_analytics` command. That is why I wanted to test the code on my sandbox and on portal-dev.conp.ca

@tkkuehn want to take a look at the code though?

app/models.py
```diff
@@ -417,3 +417,53 @@ class MatomoDailyGetSiteSearchKeywords(db.Model):
...
+    timestamp    = timestamp associated to the count
+    count        = number of clones of the GitHub repository for a given date
+    unique_count = number of unique clones for a given date
+    """
+
+    __tablename__ = "github_daily_clones_count"
+
+    id = db.Column(db.Integer, primary_key=True, autoincrement=True)
+    repo = db.Column(db.String(256))
+    date = db.Column(db.String(12))
```
tkkuehn, line 438: Why not make the `date` and `timestamp` columns of the new models with types `db.Date` and `db.DateTime`? Probably doesn't make a huge difference and would take some conversion in `_get_repo_analytics`, though, so up to you.

app/cli.py
```diff
@@ -653,3 +655,153 @@ def determine_dates_to_query_on_matomo(dates_in_database):
...
+        # get the list of dates already in the database for this repo
+        dates_in_db_dict = {}
+        db_clones_results = db.session.query(GithubDailyClonesCount.date).filter_by(repo=sub_repo).all()
+        dates_in_db_dict['clones'] = [row[0] for row in db_clones_results]
+        db_views_results = db.session.query(GithubDailyViewsCount.date).filter_by(repo=sub_repo).all()
+        dates_in_db_dict['views'] = [row[0] for row in db_views_results]
+
+        # loop through results returned by GitHub API calls and insert
+        # new data in the proper database table
+        for analytic_type in daily_stat_dict.keys():
```
tkkuehn, line 716: This can just be `for analytic_type in daily_stat_dict:`

app/cli.py
```diff
@@ -653,3 +655,153 @@ def determine_dates_to_query_on_matomo(dates_in_database):
...
+                    # for that repo in the database table
+                    continue
+
+                analytics_summary = None
+                if analytic_type == 'clones':
+                    analytics_summary = GithubDailyClonesCount()
+                elif analytic_type == 'views':
+                    analytics_summary = GithubDailyViewsCount()
+
+                if analytics_summary:
```
tkkuehn, line 732: Maybe log a warning that the `daily_stat_dict` was malformed in an `else` block after this.

tkkuehn (changes requested):
Looks good! One little logging nitpick and a question about the DB model, but this is basically good to go.

app/models.py
cmadjar: I totally agree with you on that one. Unfortunately, I tried that at the time I designed the database tables for the Matomo analytics and there were issues configuring the date field as a date. I don't remember the specific issues I had but it was impossible to get the date being of type date. I think it was with SQLite that it failed but I don't remember very well. I just remember it did not work, haha.

tkkuehn: Fair enough, that's a good reason!

cmadjar:
@tkkuehn thank you for the review. I applied the changes you mentioned. :-)

PR #531, opened 2022-04-28
Dataset card slight redesign

cmadjar:
This integrates the changes requested during the CONP dev call of 2022-04-27. This included:
- [x] remove the "Browse on GitHub" and "Source" links from the dataset card
- [x] addition of "Browse on GitHub" and "Source" links on the detailed information section of the dataset page
- [x] bolded the PROCESS and DOWNLOAD column title in the card (a redesign will be discussed with Derek later on regarding these)

Also fixed the padding after the dataset title on the card as it could overlap with the CBRAIN button when the title is very long.

Before:
Screen Shot 2022-04-28 at 10 43 34 AM

After:
Screen Shot 2022-04-28 at 10 43 41 AM


jbpoline:
looks good @cmadjar  ! One tiny improvement would be to actually move the metadata download on the dataset page and make the button "Download Dataset" 
I suppose this may be more a matter of taste but it would simplify and make things even clearer

tkkuehn (approved):
The code looks good to me and this seems to function locally!

PR #532, opened 2022-04-29
Fix labels of dataset views chart to be dataset titles instead of dataset GitHub repo name

cmadjar:
This fixes the labels used in the dataset page views analytic chart so that it displays the name of the dataset instead of the name of the GitHub repository of the dataset.

Before:
Screen Shot 2022-04-29 at 11 54 36 AM

After:
Screen Shot 2022-04-29 at 11 55 11 AM


cmadjar:
@tkkuehn if you have some time to review/test this PR, that would be awesome. Thank you! :-)

app/analytics/routes.py
```diff
@@ -107,8 +107,11 @@ def datasets_views():
                     v.nb_uniq_visitors if v.nb_uniq_visitors is not None else 0)
 
         if not exists and v.dataset_id is not None:
+            dataset_info = Dataset.query.filter_by(dataset_id=v.dataset_id).first()
+            print(dataset_info)
```
tkkuehn, line 111: We probably don't want this print statement

app/analytics/routes.py
```diff
@@ -107,8 +107,11 @@ def datasets_views():
                     v.nb_uniq_visitors if v.nb_uniq_visitors is not None else 0)
 
         if not exists and v.dataset_id is not None:
+            dataset_info = Dataset.query.filter_by(dataset_id=v.dataset_id).first()
```
tkkuehn, line 110: Nitpick: I would use `one()` instead of `first()` because `Dataset.dataset_id` is unique.

tkkuehn (changes requested):
A couple of nitpicks but this overall looks good (these charts don't show up locally for me even on master so I can't test properly but the screenshots attached to the PR look good).

app/analytics/routes.py
cmadjar: oooooops

app/analytics/routes.py
cmadjar: Thank you @tkkuehn! This has been fixed in the last commit.

tkkuehn (approved):

PR #534, opened 2022-05-02
Escape apostrophes in dataset titles

cmadjar:
https://github.com/CONP-PCNO/conp-portal/pull/525 fixed the apostrophes not being escaped in the search dataset page but that bug fix was not included for the detailed dataset page. This resolves this so now, datasets with apostrophes will render properly.


cmadjar:
@tkkuehn If you have some time :-)

tkkuehn (approved):
Looks good to me and works on my end!

PR #539, opened 2022-05-11
Do not display current month in "Number of Unique Visitors" analytics graph

cmadjar:
As discussed during the CONP dev call of May 11th, 2022, the current month is removed from the "Number of Unique Visitors" graph since the month is not over and it gives a misrepresentation of the monthly visits.

Before:
Screen Shot 2022-05-11 at 5 30 33 PM

After:
Screen Shot 2022-05-11 at 5 30 02 PM

### DIFF
--- a/app/analytics/routes.py
+++ b/app/analytics/routes.py
@@ -7,7 +7,7 @@
 import json
 import re
 
-from datetime import datetime
+from datetime import date
 from flask import render_template, request
 from flask_login import current_user
 from app.analytics import analytics_bp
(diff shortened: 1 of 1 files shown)


cmadjar:
@tkkuehn feel like reviewing and testing this?

tkkuehn (approved):
Looks reasonable to me!

Nitpicks that you don't have to implement (a little more readable imo but same function):

- We could use `datetime.date` instead of `datetime.datetime`.
- We could directly compare the year and month instead of converting to strings.

Something like:

```
current_date = date.today()
for v in daily_visits:
    visit_date = date.fromisoformat(v.date)
    if (visit_date.year == current_date.year) and (visit_date.month == current_date.month):
        continue
...
```

tkkuehn:
I should also note that this chart never worked locally for me so I can't test, but the change is straightforward and your screenshot looks good.

cmadjar:
@tkkuehn thank you for the suggestion! I applied the changes.

PR #545, opened 2022-05-18
Fix pipeline data in the dashboard graph

cmadjar:
This fixes the data being displayed in the bar graph for the number of pipelines. 

Before: 
Screen Shot 2022-05-18 at 3 00 54 PM

After:
Screen Shot 2022-05-18 at 3 01 22 PM

Note: the graph for the pipeline was also reconstructed on the same data with Python for double checks:
Screen Shot 2022-05-18 at 2 38 04 PM


cmadjar:
Closed since the changes were integrated in #546

PR #551, opened 2022-06-03
Sort list of pipelines/datasets for CBRAIN modal

tkkuehn:
## Checklist

- [ ] **DO** Unit tests pass.

## Purpose
This PR sorts the lists of pipelines or datasets that are presented in the "Process on CBRAIN" modal.

## Current behaviour
The pipelines/datasets are shown in default sort order (I think by number of downloads, descending), which is unpredictable and does not help identify the one you want.

## New behaviour
The pipelines/datasets are sorted by title, ascending (other than "None"), organizing them in a predictable way.
 
- [ ] Yes
- [ x ] No


cmadjar (changes requested):
@tkkuehn I just tried it locally and it does not seem to be ordering the list of pipelines or datasets alphabetically. See screenshot below:



Screen Shot 2022-06-08 at 10 57 09 AM

tkkuehn:
That's odd, I checked again and it works for me.

That said, the screenshots you attached don't show the part of the portal this is meant to change. Just to make sure it's clear, this is supposed to sort the list of datasets/pipelines in the modal (i.e. pop-up) that shows when you click the "Process On CBRAIN" button.

This is what I see locally -- If the choices in the pop-up aren't sorted, there's definitely a bug I'll try to track down:
Screenshot from 2022-06-08 13-12-00
Screenshot from 2022-06-08 13-11-49

cmadjar (approved):

PR #556, opened 2022-09-09
CBRAIN redirect URL

tkkuehn:
---
name: CBRAIN redirect URL
about: Set up a URL to redirect to CBRAIN, allowing those redirects to be logged.
title: 'CBRAIN redirect URL'
labels: enhancement
assignees: ''

---

## Checklist

- [ ] **DO** Unit tests pass.

The unit tests fail here and on master, I think in a way related to the addition of ARK IDs.

## Purpose

This adds a URL, `/cbrainredirect?cbrainurl={url}`, that redirects to a CBRAIN URL. The idea here is that this will allow Matomo (and/or Apache and/or Flask?) to log those redirects to CBRAIN as a metric. My knowledge of the deployment details is fuzzy, so someone who knows more should confirm whether this is sufficient or whether I should add a new database table to count these redirects or something else.

## Current behaviour

Users are linked directly to CBRAIN, and it's not possible to keep track of how often those links are followed.

## New behaviour

Users are linked to CBRAIN through a redirect URL that we can log.
 
- [ ] Yes
- [x] No

## Implementation Detail

The dataset and pipeline search APIs modify the CBRAIN URLs to place them in the `/cbrainredirect` path. Minor changes to the CBRAIN modal React components accomodate this change.


emmetaobrien (approved):
Seems fine to me.

PR #579, opened 2023-05-10
Remove "dark mode" changes from analytics

tkkuehn:
The analytics charts respect browsers' "dark mode" settings, unlike anything else on the portal, which ends up making the charts totally illegible.

This is a somewhat hacky fix, which takes the version of https://code.highcharts.com/css/highcharts.css that I retrieved today, and removes the following block that changes the CSS based on the browser's dark mode settings:

```
@media (prefers-color-scheme: dark) {
    :root {
        /* UI colors */
        --highcharts-background-color: rgb(48, 48, 48);

        /*
            Neutral color variations
            https://www.highcharts.com/samples/highcharts/css/palette-helper
        */
        --highcharts-neutral-color-100: rgb(255, 255, 255);
        --highcharts-neutral-color-80: rgb(214, 214, 214);
        --highcharts-neutral-color-60: rgb(173, 173, 173);
        --highcharts-neutral-color-40: rgb(133, 133, 133);
        --highcharts-neutral-color-20: rgb(92, 92, 92);
        --highcharts-neutral-color-10: rgb(71, 71, 71);
        --highcharts-neutral-color-5: rgb(61, 61, 61);
        --highcharts-neutral-color-3: rgb(57, 57, 57);

        /* Highlight color variations */
        --highcharts-highlight-color-100: rgb(122, 167, 255);
        --highcharts-highlight-color-80: rgb(108, 144, 214);
        --highcharts-highlight-color-60: rgb(94, 121, 173);
        --highcharts-highlight-color-20: rgb(65, 74, 92);
        --highcharts-highlight-color-10: rgb(58, 63, 71);
    }
}
```

The analytics page now uses this modified CSS file instead of downloading Highcharts' copy, resolving the problem.

### DIFF
--- a/app/static/css/highcharts.css
+++ b/app/static/css/highcharts.css
@@ -1,0 +1,1178 @@
+/**
+ * @license Highcharts
+ *
+ * (c) 2009-2022 Torstein Honsi
+ *
+ * License: www.highcharts.com/license
+ */
+
+:root,
+.highcharts-light {
+    /* Colors for data series and points */
+    --highcharts-color-0: #2caffe;
+    --highcharts-color-1: #544fc5;
+    --highcharts-color-2: #00e272;
+    --highcharts-color-3: #fe6a35;
+    --highcharts-color-4: #6b8abc;
+    --highcharts-color-5: #d568fb;
+    --highcharts-color-6: #2ee0ca;
+    --highcharts-color-7: #fa4b42;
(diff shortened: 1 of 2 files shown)


emmetaobrien:
As an immediate fix this seems reasonable to me; I think we should consider how the portal works for people who need to use dark-mode or equivalent generally, on accessibility grounds, going forward.

PR #582, opened 2023-05-30
Make dataset search a little smarter

tkkuehn:
## Purpose
Makes dataset search a little more precise.

## Current behaviour
Dataset search looks (case-insensitively) for the exact search term anywhere in the DATS.json (including in its keys).

## New behaviour
Dataset search looks (case-insensitively) for the exact search term anywhere in a select group of parsed DATS.json values and the dataset README. Note: this is only a minor improvement, but was relatively easy to implement. Further improvements could include splitting search terms that aren't wrapped by quotation marks, adding boolean operators, potentially others? That will add a maintenance burden but this is an improvement at a low investment of dev time.

### DIFF
--- a/app/search/models.py
+++ b/app/search/models.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 import datetime as dt
 from functools import lru_cache
 import os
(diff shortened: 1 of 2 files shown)


tkkuehn:
@GHPBZ @emmetaobrien I don't have permissions to request review on this repo, but could you take a look at this?

emmetaobrien:
That certainly looks a worthwhile improvement to me.

PR #585, opened 2023-05-31
Force light background on front-page charts

tkkuehn:
This extends the changes in #579 to the front page, which I missed the first time around.


emmetaobrien (approved):
lgtm

